PR #74, opened 2014-06-02
Adding an optional directory argument to the CLI for the check command

pcorliss:
Not sure if the specs meet with your expectations. Let me know if you'd like me to rework them in a particular way.


postmodern:
Oops, forgot about this entirely but ended up adding it in be9c5c5dfb5a2d80ecfc26395bc6946e1711cb87.

PR #86, opened 2015-02-16
print explanatory message and exit if $HOME is nil.

heartpunk:


tehgeekmeister:
Lost a solid couple hours to this, would love to know that no one else will have to. A better fix would include a way to set where to put/look for the advisory database, but I think it's better to start the conversation with the minimal fix.

postmodern:
bundler-audit 0.8.0 now uses `Gem.user_home` which tries to auto-detect the home directory if `HOME` isn't set.

PR #89, opened 2015-03-31
Add option to update before running check

adrianmacneil:
I've integrated bundler-audit into our CI (it's fantastic btw, thanks for all the work on this).

However, it's difficult to run more than one command at a time. Right now I'm solving this with a script which runs `bundle-audit update && bundle-audit check`, but I figured it would make sense to natively support the option to update the database and run a check in a single command.

Let me know what you think.


infertux:
:+1: good idea, I resort to the same `bundle-audit update && bundle-audit check` "hack" in my CI at the moment.

postmodern:
Good idea. Travis only crashed because the i18n gem now requires >= 1.9.3.

PR #90, opened 2015-04-22
Treat internal git/http hosts as secure

### LINKED ISSUE #72: Whitelist insecure git sources if they are internal IPs
Resolve the host of insecure git sources and check if they are non-routable IPs.

sds:
Many organizations have references to internal git repositories in their
`Gemfile`s. Allow these repositories to be treated as safe since they
are resolvable only internally, and are controlled by the organization.

Closes #72

### DIFF
--- a/lib/bundler/audit/scanner.rb
+++ b/lib/bundler/audit/scanner.rb
@@ -2,7 +2,10 @@ require 'bundler'
 require 'bundler/audit/database'
 require 'bundler/lockfile_parser'
 
+require 'ipaddr'
+require 'resolv'
 require 'set'
+require 'uri'
 
 module Bundler
   module Audit
@@ -70,6 +73,7 @@ module Bundler
           when Source::Git
             case source.uri
             when /^git:/, /^http:/
+              next if internal_host?(source.uri)
               yield InsecureSource.new(source.uri)
             end
           when Source::Rubygems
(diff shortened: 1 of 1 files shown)


sds:
The build failures are related to installing the `i18n` gem on EOLed versions of Ruby, and are unrelated to this change.

retornam:
@sds please  rebase your commit to kick off the travis test runs again. The i18n gem issue has been fixed.

sds:
@retornam thanks for the response. I've rebased onto the latest `master` at time of writing (02de26e7c8c8194500dbddaed45c5ad8d1619bf1).

sds:
Incorporated feedback from @postmodern in d414a12ad85355ff15204a64be0bebe0d5096842, namely:

* Split out `internal_host?` helper from `internal_ip?` to separate URI host lookup from IP subnet-checking logic.
* Check all addresses returned `Resolv.getaddresses` instead of just the first address returned by `Resolv.getaddress`

I also added support for IPv6 private addresses.

postmodern:
An aside, I really wish IPAddress provided a method for determining if an address is from one of the RFC1918 ranges. I recently had to build a list of non-routable address ranges recently.

bhh:
as it seems people are discussing that topic already in the ipaddress project. https://github.com/bluemonk/ipaddress/issues/54

PR #94, opened 2015-07-02
Bug ignoring CVE-2015-3448

tapajos:
# The problem

I've found a bug ignoring `CVE-2015-3448`.

```
$ bundle exec bundle-audit check --ignore CVE-2015-3448 CVE-2015-1820 CVE-2015-3227
Name: rest-client
Version: 1.6.9
Advisory: CVE-2015-3448
Criticality: Unknown
URL: http://www.osvdb.org/show/osvdb/117461
Title: Rest-Client Gem for Ruby logs password information in plaintext
Solution: upgrade to >= 1.7.3

Vulnerabilities found!
```

Debugging it I figured out that the advisory ID is not `CVE-2015-3448` but `OSVDB-117461`.

```
#<struct Bundler::Audit::Advisory
 path="/Users/tapajos/.local/share/ruby-advisory-db/gems/rest-client/OSVDB-117461.yml",
 id="OSVDB-117461",
 url="http://www.osvdb.org/show/osvdb/117461",
 title="Rest-Client Gem for Ruby logs password information in plaintext",
 description="Rest-Client Ruby Gem contains a flaw that is due to the application logging password information in plaintext. This may allow a local attacker to gain access to password information.",
 cvss_v2=nil,
 cve="2015-3448",
 osvdb=117461,
 unaffected_versions=[],
 patched_versions=[Gem::Requirement.new([">= 1.7.3"])]>
```

CLI is printing the Advisory based on the code bellow but it seems to be wrong because (looks like) an advisory can be `advisory.cve` and `advisory.osvdb` at the same time.

``` ruby
if advisory.cve
    say "CVE-#{advisory.cve}"
elsif advisory.osvdb
    say advisory.osvdb
end
```
# The fix

I put a really simple fix printing the `advisory.id` (what we need to add in ignore list)

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -91,11 +91,7 @@ module Bundler
 
         say "Advisory: ", :red
 
-        if advisory.cve
-          say "CVE-#{advisory.cve}"
-        elsif advisory.osvdb
-          say advisory.osvdb
-        end
+        say advisory.id
 
         say "Criticality: ", :red
         case advisory.criticality
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -22,7 +22,7 @@ describe "CLI" do
     it "should print advisory information for the vulnerable gems" do
       advisory_pattern = /(Name: [^\n]+
 Version: \d+.\d+.\d+
-Advisory: CVE-[0-9]{4}-[0-9]{4}
+Advisory: (OSVDB-[0-9]{5,6}|CVE-[0-9]{4}-[0-9]{4})
 Criticality: (High|Medium)
 URL: http:\/\/(direct|www\.)?osvdb.org\/show\/osvdb\/\d+
 Title: [^\n]*?


tapajos:
I'm having some issues running the tests and I didn't figure out why.

Probably this change requires test changes and I'm not ignoring it, I just want to share it and hear some feedback but I will work to have all tests passing.

tapajos:
tests passing.

I figured out my issue, the problem was with data/ruby-advisory-db. I cloned the last one and it worked but looks like I'm doing something wrong. Would be nice have it in README.

postmodern:
Seems like the problem is we should allow ignoring based on OSVDB or CVE. Additionally, we should print both CVE and OSVDB, not just what the primary ID of the advisory file is.

tapajos:
I agree.

I will work on it.

postmodern:
I think that code should work, although it needs specs. Also, not sure how to best print both CVE and/or OSVDB.

pjscopeland:
I'm a bit of a noob at this, and this still wasn't working for me:

```
+ bundle audit check --ignore=126329 126330 126331
[Other vulnerabilities]

Name: sidekiq-pro
Version: 1.6.0
Advisory: 126329
Criticality: Unknown
URL: https://github.com/mperham/sidekiq/commit/a695ff347ae50f641dfc35189131b232ea0aa1db
Title: Sidekiq Pro Gem for Ruby web/views/batch.erb Class and ErrorMessage Elements
Reflected XSS
Solution: upgrade to >= 2.0.2

Name: sidekiq-pro
Version: 1.6.0
Advisory: 126330
Criticality: Unknown
URL: https://github.com/mperham/sidekiq/commit/99b12fb50fe244c5a317f03f1bed9b333ec56ebe
Title: Sidekiq Pro Gem for Ruby web/views/batch{,es}.erb Description Element XSS
Solution: upgrade to >= 1.9.1

Name: sidekiq-pro
Version: 1.6.0
Advisory: 126331
Criticality: Unknown
URL: https://github.com/mperham/sidekiq/commit/651400ed8f237118346895c99dc28ca94f3169d3
Title: Sidekiq Pro Gem for Ruby CSRF in Job Filtering
Solution: upgrade to >= 2.0.6

Vulnerabilities found!
```

On the other hand, prepending the advisories with `OSVDB-` _did,_ but I needed to come to this page to figure it out:

```
+ bundle audit check --ignore=OSVDB-126329 OSVDB-126330 OSVDB-126331
[Other vulnerabilities]

Vulnerabilities found!
```

tapajos:
Probably worth closing this. I'm not working with ruby anymore

duhaime:
Hi @tapajos @postmodern why was this closed? This still seems to be a problem

tapajos:
> Hi @tapajos @postmodern why was this closed? This still seems to be a problem

I've closed this because I will not have time to improve ir.

duhaime:
The issue itself remains, so I think the issue should remain open (someone else from the community can help close it)

tapajos:
I think the repo owner can reopen it.--
Marcos Tapajós
***@***.***
Mobile: +55 21 98709-8693
Skype: marcostapajos
Gtalk/MSN: ***@***.***

postmodern:
The code has changed since then. The advisory ID is now printed in [`lib/bundler/audit/cli/formats/text.rb`](https://github.com/rubysec/bundler-audit/blob/c1d2fedd66d62275162c36d97115f5a484856e54/lib/bundler/audit/cli/formats/text.rb#L72-L80). Instead of just printing `advisory.id` the code highlights `CVE` or `GHSA` as red. Advisories can have a CVE but no GHSA, a GHSA but no CVE, or have both a CVE and a GHSA. We've also deprecated `OSVDB` IDs from bundler-audit. Also, the scanner code will now ignore advisories if the given advisory ID matches _any_ of the advisory [identifiers](https://github.com/rubysec/bundler-audit/blob/c1d2fedd66d62275162c36d97115f5a484856e54/lib/bundler/audit/advisory.rb#L88-L130). I don't think this is still an issue? Could you confirm with a simple Steps To Reproduce example?

duhaime:
Thanks for your follow up. When I attempted to ignore an issue using the CVE ID (both with and without the `CVE-` prefix) the ignore was not successful--nothing happened and the error continued to display when I re-ran the audit. Possibly we need to upgrade? We're on bundler-audit `0.9.1`.

postmodern:
@duhaime could you paste the command you are using? Testing locally against a test rails app `bundle-audit check --ignore CVE-YYYY-NNNN` does ignore that CVE. I did notice that Thor does weird things with array type options, `--ignore FOO --ignore BAR` will actually override the first `FOO`. To pass in multiple values you have to use `--ignore FOO BAR`, which isn't how POSIX option parsing normally works and can be confusing.

duhaime:
Hmm, very probably this is user error. We had ~6 or so different CVEs that sprang. We wanted to ignore them all so I tried ignoring the first `bundle-audit check --ignore CVE-2023-22792`. I did the same for the next first item, and so on, until I realized that I was trying to ignore an item I had already tried to ignore (e.g. that my first attempt had failed). 

Does one need to ignore all the CVEs at once?

postmodern:
@duhaime yes or add them to the [`.bundler-audit.yml`](https://github.com/rubysec/bundler-audit#configuration-file) configuration file.

PR #96, opened 2015-07-09
better non CLI usage

bkarstaedt:
Especially if you want to audit many repositories at once you want to have more control over the auditing process and the output usage. In this PR I'd like to encourage the reader to use bundler-audit programmatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,22 @@ Ignore specific advisories:
 
     $ gem install bundler-audit
 
+## Non-CLI Usage
+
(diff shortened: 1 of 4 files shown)


postmodern:
By adding `#to_s` methods that format the results into CLI output, you're kind of creating a CLI interface ;)

bundler-audit is supposed to be a CLI util, so not sure whether I want to support non-CLI usage; although we do provide a rake task for automation purposes.

PR #100, opened 2015-07-24
Add --quiet as a command line option for update

### LINKED ISSUE #78: Update databse using `--quiet` flag during git clone
Hey,

It's will be great to add optional `--quiet` flat during `git clone https://github.com/rubysec/ruby-advisory-db.git`.

We use `bundle-audit` during CI build and just don't want to pollute output log.

latortuga:
Fixes #78, simply pass through the command line option --quiet to git commands.

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -33,9 +33,10 @@ module Bundler
       method_option :verbose, :type => :boolean, :aliases => '-v'
       method_option :ignore, :type => :array, :aliases => '-i'
       method_option :update, :type => :boolean, :aliases => '-u'
+      method_option :quiet, :type => :boolean, :aliases => '-q', default: false
 
       def check
-        update if options[:update]
+        update(options[:quiet]) if options[:update]
 
         scanner    = Scanner.new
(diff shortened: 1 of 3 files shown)


lib/bundler/audit/cli.rb
```diff
@@ -60,10 +61,10 @@ def check
       end
 
       desc 'update', 'Updates the ruby-advisory-db'
-      def update
+      def update(quiet = false)
```
latortuga: In my own projects, I would normally make this a keyword arg but the README says you are trying to support Ruby 1.9.3 which doesn't have them. Thus this is what I went with.

postmodern: Could you rewrite this to use `options={}` style of passing in a `quiet: true` option?

latortuga: Yes although I would err on the side of YAGNI. It's the only option `update` accepts right now. If you really want it, I will change it.

postmodern: We may need to add other options in the future. I like to avoid changing argument types, as that technically changes the API.

Issue #106, opened 2015-08-17
InsecureSource results in exit code 1


krigar:
We're about to integrate bundler-audit into our CircleCI build process and it's looking really promising except for one fact: ignoring vulnerabilities leads to none being listed by bundler-audit, but it still exits with `exit code 1`, making CircleCI think it failed.

The reason why we ignore some vulnerabilities is because we're running on a forked version of https://github.com/spree/spree and we have to monkey patch their security patches instead of upgrading the version.

Example output:
```
$ bundle exec bundle-audit check --update --ignore OSVDB-119205 OSVDB-125699 OSVDB-125701
Updating ruby-advisory-db ...
From https://github.com/rubysec/ruby-advisory-db
 * branch            master     -> FETCH_HEAD
Already up-to-date.
ruby-advisory-db: 226 advisories
Vulnerabilities found!
```

krigar:
Figured this one out, it was due to some git repositories in our lockfile that got marked as `InsecureSource`.

I understand the need to generate a warning when using external git repositories, but should the warning result in `vulnerable` being set to `true`?

postmodern:
If the repositories are insecure, an attacker could MITM them and inject arbitrary code. If your threat model involves MITMing, then it makes sense to mark them as vulnerable; plus it's not hard to switch to `https://` or `git@`.

reedloden:
Why would `git@` be more secure than `http://`? Aren't they effectively both unauthenticated and unencrypted? Are you referring to git over ssh, perhaps?

Mange:
Yes. `git@` is the syntax for GitHub as the SSH user is named `git` there.
`git://` and `http://` are insecure.

lör 29 aug 2015 00:29 Reed Loden <notifications@github.com> skrev:

reedloden:
Oh, duh. Reading comprehension fail. My bad.

JuanitoFatas:
@krigar Hi! Did you solve the rake task? You could use the rake task in your project from #115 or try [this commit](https://github.com/thoughtbot/hound/commit/59663c928fe8fef4971688ea8c1cf52a68c8f6a4). 

Then I think this issue can be closed? InsecureSource results in exit code 1 is correct as @postmodern mentioned in [this comment](https://github.com/rubysec/bundler-audit/issues/106#issuecomment-135891016).

Thanks!

postmodern:
Can this be closed?

krigar:
Hey, yeah, I'll close it now.

PR #107, opened 2015-08-17
Include an ignore-insecure-sources flag

krigar:
As mentioned here rubysec/bundler-audit#106

Currently, warnings about insecure sources - such as external git or http URIs - marks the lockfile as vulnerable. This leads to the gem exiting with code 1, effectively raising an error in the CI build process.

These warnings are entirely valid, but there should be a way of ignoring them.

### DIFF
--- a/lib/bundler/audit/scanner.rb
+++ b/lib/bundler/audit/scanner.rb
@@ -68,7 +68,7 @@ module Bundler
         ignore = Set[]
         ignore += options[:ignore] if options[:ignore]
 
-        scan_sources(options,&block) unless options[:ignore_insecure_sources]
+        scan_sources(options,&block)
         scan_specs(options,&block)
 
         return self
--- a/lib/bundler/audit/scanner.rb
+++ b/lib/bundler/audit/scanner.rb
@@ -68,7 +68,7 @@ module Bundler
         ignore = Set[]
         ignore += options[:ignore] if options[:ignore]
 
-        scan_sources(options,&block)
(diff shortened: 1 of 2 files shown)


gshaw:
:+1:

lib/bundler/audit/scanner.rb
```diff
@@ -68,7 +68,7 @@ def scan(options={},&block)
         ignore = Set[]
         ignore += options[:ignore] if options[:ignore]
 
-        scan_sources(options,&block)
+        scan_sources(options,&block) unless options[:ignore_insecure_sources]
```
reedloden: Wouldn't it be better to still scan and output any problems, but just not consider it a failure?

krigar: That was actually my first solution, but wasn't sure if it would be confusing. Can easily do something like this: https://gist.github.com/krigar/a48e505a9bd5f3aa5d8a

I'll change the pull request to adopt that approach instead.

postmodern:
Is there a reason one cannot use `https://` or `git@` sources/repos?

gshaw:
We currently use http:// in our continuous build environment so that we can cache gems with Squid to significantly reduce the burden we place on the gem source servers. The build environment itself is secure enough that we feel it unlikely to be susceptible to MitM or DNS spoofing attacks.

postmodern:
Hmm, now that I think about it. The `--ignore-insecure-sources` option should whitelist specific URIs.

PR #109, opened 2015-09-01
Move file reading responsibility to the CLI

stevenocchipinti:
This means `Scanner` is a simple wrapper for `LockfileParser` and accepts
the same arguments (content in the form of a string, not a directory path).

This also means that `Scanner` is able to be called with content from
anywhere (for example, github's content API) instead of a hardcoded file
path and makes it more useful as a library other projects can depend on.

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -34,7 +34,8 @@ module Bundler
       method_option :ignore, :type => :array, :aliases => '-i'
 
       def check
-        scanner    = Scanner.new
+        root = File.expand_path(Dir.pwd)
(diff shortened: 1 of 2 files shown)


postmodern:
Sounds reasonable. Perhaps we should add `Scanner.parse` and `Scanner.open`?

stevenocchipinti:
Could do that. So just to clarify, would you imagine moving the file reading logic back in to `Scanner` and allow either of these invocations: 

```
scanner = Scanner.open('Gemfile.lock')
```
and
```
scanner = Scanner.parse(File.read(File.join(root, 'Gemfile.lock')))
```

Would you leave this implementation of `#initialize`?

PR #110, opened 2015-09-02
Support checking for newer gems.locked in addition to Gemfile.lock

skorth:
Its just for discussion cause i didn't get the tests running. "gems.locked" is supported since Bundler 1.8.0 and will be default in 2.0 (as far as i know).

### DIFF
--- a/lib/bundler/audit/scanner.rb
+++ b/lib/bundler/audit/scanner.rb
@@ -39,8 +39,15 @@ module Bundler
       def initialize(root=Dir.pwd)
         @root     = File.expand_path(root)
         @database = Database.new
+
+        @file = if File.exists? File.join(@root,'Gemfile.lock')
+          File.join(@root,'Gemfile.lock')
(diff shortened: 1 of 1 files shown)


jlecour:
I have a few projects where I've switched to `gems.rb`/`gems.locked`. It would be great to have bundler-audit work with them too.

BTW, Thanks @postmodern for this great tool.

postmodern:
The 0.8.0 branch supports a `--gemfile-lock` option. After 0.8.0, we should probably start checking for `gems.locked` as well as `Gemfile.lock`. Would also prefer the logic gets DRYed up by using a constant Array of Gemfile-lock names.

postmodern:
I went ahead and added auto-detection code in 46b249a if no explicit lock-file name is given. Double check my logic.

I also tested whether bundler 2.3.6 would pick `gems.locked` over `Gemfile.lock`, but it looks like it still prefers `Gemfile.lock` over `gems.locked`.

PR #119, opened 2015-11-03
Update command examples and badges in the readme

tas50:


postmodern:
Implemented in 42e992a2efb5311978c0e0fd612517cab1621bac. That CLI example is meant to show that `bundle-audit check` and also update prior to auditing.

PR #121, opened 2015-11-05
Update fails with non-zero exit instead of silent failure

### LINKED ISSUE #114: Exits with `0` for `bundle-audit update` network failure, expected non-zero exit code
When running `bundle-audit update` is it expected to return a non-zero exit code when the update server is unavailable? Currently I'm seeing a zero exit code (bundler-audit 0.4.0). 

I tried testing this by disabling my network connection and running:

```bash
$ bundle-audit --version
bundle-audit 0.4.0 (advisories: 233)

$ bundle-audit update
Updating ruby-advisory-db ...
fatal: unable to access 'https://github.com/rubysec/ruby-advisory-db.git/': Could not resolve host: github.com
ruby-advisory-db: 233 advisories

# zero exit code, expected non-zero:
$ echo $?
0
```
A reason for preferring a non-zero exit code instead, is for environments that are running unattended jobs like:

```bash
# If advisory db fails to update, may not have latest vulnerabilties:
bundle-audit update && bundle-audit
```
These environments could report that the `bundle-audit` is good (no vulns. found) when the advisory db is out-of-date. `bundle-audit update` may persistently fail due to restrictive firewall policies, meaning the advisory db is never updated successfully.

eliotsykes:
Fixes #114. Still needs specs. I'm happy to add specs, please just give me a sign that there's the smallest possibility of this being a welcome change, then I'll go ahead and write specs.

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -63,7 +63,13 @@ module Bundler
       def update
         say "Updating ruby-advisory-db ..."
 
-        Database.update!
+        if Database.update!
+          say "Updated ruby-advisory-db", :green
+        else
+          say "Failed updating ruby-advisory-db!", :red
+          exit 1
+        end
+
         puts "ruby-advisory-db: #{Database.new.size} advisories"
       end
 


tute:
I'd like this merged. I'd rather have this gem be explicit about the error, than route around it, like we are discussing in https://github.com/thoughtbot/suspenders/pull/651. Thank you!

postmodern:
Looks good. Not sure how aggressive the CLI specs should be.

tute:
Thanks so much!

tute:
@eliotsykes do you plan to make a release with this fix? Thanks!

eliotsykes:
Hi @tute, I'm not able to, this was a drive-by PR :-)

PR #124, opened 2015-11-13
Allow whitelisting insecure sources

zapo:
This is a follow up on #107, pretty similar and highly inspired from it, but instead of being a flag this allows explicitly whitelisting sources as suggested in the last comment of the initial PR. I figured since their were no activity since August so I could create this one.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -40,7 +40,7 @@ namespace :spec do
   task :bundle do
     root = 'spec/bundle'
 
-    %w[secure unpatched_gems insecure_sources].each do |bundle|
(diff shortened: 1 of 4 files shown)


AlexWayfer:
> @zapo closed this 10 days ago

Why?

zapo:
@AlexWayfer it's been sitting here for 5 years without much response from maintainers. I was doing some summer cleanup and realized this conflicts with main trunk and that I wouldn't plan on updating it.

AlexWayfer:
OK, I just don't understand such "cleanups". I can have a hungered of open PRs and they're not bothering me.

PR #126, opened 2015-11-18
Prepend "OSVDB-" prefix in cli output for advisories

### LINKED ISSUE #125: Can full Advisory ID be displayed in the output?
Was having a play around and wondering why the following command was not ignoring the given vulnerability:

``` shell
$ be bundle-audit check --ignore 126747
Name: uglifier
Version: 2.7.1
Advisory: 126747
Criticality: Unknown
URL: https://github.com/mishoo/UglifyJS2/issues/751
Title: uglifier incorrectly handles non-boolean comparisons during minification
Solution: upgrade to >= 2.7.2
```

After doing some debugging, I found that the advisory property being used to compare to the ignore list was the `id` property ("OSVDB-126747"), but the code prints the `osvdb` property ("126747") which did not include the prefix.

``` ruby
unless ignore.include?(advisory.id)
  yield UnpatchedGem.new(gem,advisory)
end
```

``` ruby
        say "Advisory: ", :red

        if advisory.cve
          say "CVE-#{advisory.cve}"
        elsif advisory.osvdb
          say advisory.osvdb
        end
```

Is there a reason for not displaying the full Advisory ID in the output?

reedloden:
Fixes #125.

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -101,9 +101,9 @@ module Bundler
         say "Advisory: ", :red
 
         if advisory.cve
-          say "CVE-#{advisory.cve}"
+          say advisory.cve_id
         elsif advisory.osvdb
-          say advisory.osvdb
+          say advisory.osvdb_id
         end
 
         say "Criticality: ", :red


bethesque:
Is there a reason not to just use `advisory.id`? Seeing as that's the property that's being compared to the ignore list.

Davidslv:
@bethesque not anymore, since https://github.com/rubysec/bundler-audit/commit/4e32fca89d75f0e249671431ff38aadc02bfb28b

reedloden:
Included in another merged PR, so closing.

PR #129, opened 2015-11-21
CLI update test coverage

eliotsykes:
Providing test coverage for #121 / https://github.com/rubysec/bundler-audit/pull/121#issuecomment-156596349

### DIFF
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -78,4 +78,26 @@ Insecure Source URI found: http://rubygems.org/
       expect(subject.strip).to eq("No vulnerabilities found")
     end
   end
+
+  describe "update" do
+
+    let(:update_command) { "#{command} update" }
+    let(:bundle)         { 'secure' }
+    let(:directory)      { File.join('spec','bundle',bundle) }
+
+    subject do
+      Dir.chdir(directory) { sh(update_command) }
+    end
+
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -3,6 +3,7 @@
 
 require 'rspec'
 require 'bundler/audit/version'
+require 'bundler/audit/database'
```
eliotsykes: Running specs individually were failing without require-ing database, e.g. `bundle exec rspec spec/integration_spec.rb`

spec/integration_spec.rb
```diff
@@ -78,4 +78,26 @@
...
+    subject do
+      Dir.chdir(directory) { sh(update_command) }
+    end
+
+    context "when advisories update successfully" do
+      it "should print status" do
+        expect(subject).not_to include("Fail")
+        expect(subject).to include("Updating ruby-advisory-db ...\n")
+        expect(subject).to include("Updated ruby-advisory-db\n")
+        expect(subject.lines.to_a.last).to match(/ruby-advisory-db: [1-9]\d+ advisories/)
```
eliotsykes: Needed `to_a` to keep ruby 1.9.3 happy which returns `Enumerator` from `String#lines` (vs `Array` in Ruby 2+). `Enumerator#last` not a thing.

PR #133, opened 2015-12-23
Only try to do a git pull if ".git" exists

alex:
This allow `update` to work if the directory already exists.


alex:
Specifically this matters if you try to cache the directory on Travis, because travis will always create that directory, even on first run when its empty.

alex:
Thanks!

postmodern:
Actually should have been two if statements, one to check if the repo exists, then another to check if it's a git repo.

PR #134, opened 2016-01-05
Add a new command `list`

richo:
List is poorly named, but I couldn't come up with anything better
offhand. The usecase is something along the lines of:

        ls /opt/someapp/gems | bundler-audit list

For when you have a third party application with a bunch of bundled
gems, but the Gemfile and Gemfile.lock not included with the deployed
application.

Please feel free to propose a better name for the command!

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -83,6 +83,25 @@ module Bundler
         puts "#{File.basename($0)} #{VERSION} (advisories: #{database.size})"
       end
 
+      desc 'list', 'Processes a list of installed gems'
+      def list
+        db = Database.new
+        STDIN.each_line do |line|
+          name, version = unpack_version(line.chomp)
+          # Explicitly ignore anything with a version of "java"
+          next if version == "java"
+
+          spec = Gem::Specification.new do |s|
+            s.name = name
+            s.version = version
+          end
+
+          db.check_gem(spec) do |advisory|
+            print_advisory spec, advisory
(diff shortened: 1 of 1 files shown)


postmodern:
Maybe `batch` is a better name? What is your usecase here? Scanning the output of `gem list`?

postmodern:
Yeah, I think we can come up with a better way to check the contents of a gem-dir, probably via rubygem's own API.

richo:
`batch` works for me.

I have no idea what rubygems provides in terms of ways to do this programatically, but my usecase was literally that I unpacked a third party application, and had a bunch of `gemname-version` directories that I wanted to quickly scan.

richo:
bump, any news here?

postmodern:
Still too busy with commercial work. This now seems more like a special use case (inspecting packed apps). You should look into the rubygems API and figure out how to load `Gem::Specification` objects, which can then be ran through `Database#check_gem`. I am considering extracting the ruby-advisory-db portion of the code, for non-bundler use cases.

richo:
That's plausible, although I feel like the next time I have a variation on this problem it'll come back to the same underlying problem. I tend toward thinking a generic "Tell me if this of gems is ok" is the fundamental interface that the others should stem from, but up to yu.

postmodern:
@richo That also seems reasonable! I can change the API to accept a name/version tuple, and map the `Gem::Specification`s to that.

richo:
That would be perfect!

PR #138, opened 2016-02-26
Tick version to resolve dependency issues.

incazteca:
Version 0.4.0 was released back in June of 2015. Since then about 15
commits have been made. Even though many changes were made the version
number remained the same. This has cause some confusion as the version
0.4.0 listed on rubygems does not have features that the current version
of 0.4.0 has.

### DIFF
--- a/lib/bundler/audit/version.rb
+++ b/lib/bundler/audit/version.rb
@@ -18,6 +18,6 @@
 module Bundler
   module Audit
     # bundler-audit version
-    VERSION = '0.4.0'
+    VERSION = '0.4.1'
   end
 end


incazteca:
Personally ran into issues at work where some versions of 0.4.0 had a rake task implemented and where other versions of 0.4.0 that were sourced from rubygems did not have the task implemented causing failing CI runs.

@postmodern I personally think that the bundler-audit version should be `0.5.0`. However, I'm submitting a PR that sets it to `0.4.1` since I noticed there was a `0.5.0` branch and did not want to step on too many toes. There should also probably be a requirement that a PR tick the version number.

postmodern:
I'm going to release 0.5.0.

incazteca:
Version 0.5.0 has been released

Issue #140, opened 2016-03-02
Add support for Rails 4.2.6.*


jasnow:
Add support for Rails 4.2.6.*. Example:
```
Name: actionpack
Version: 4.2.6.rc1
Advisory: CVE-2016-2098
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/ly-IH-fxr_Q
Title: Possible remote code execution vulnerability in Action Pack
Solution: upgrade to ~> 3.2.22.2, ~> 4.1.14.2, ~> 4.2.5.2
```

postmodern:
You'll want to submit an amendment to the [ruby-advisory-db](https://github.com/rubysec/ruby-advisory-db/issues) for `gems/actionpack/CVE-2016-2098.yml`.

jasnow:
writing issue

PR #142, opened 2016-03-24
Update ruby-advisory-db module

gbmoretti:
Hey,

I just updated the ruby-advisory-db submodule. This fixes issues like #140


postmodern:
bundler-audit 0.8.0 stopped vendoring ruby-advisory-db as a git module, but instead auto-downloads it to avoid using a stale copy that was shipped with the bundler-audit gem.

PR #149, opened 2016-06-02
Run as external Bundler command

vassilevsky:
When a user runs `bundle COMMAND` and Bundler does not include COMMAND, it tries to run an executable named `bundler-COMMAND`.

This commit exploits this behavior and allows the user to run `bundle audit` to run bundler-audit.

PR #153, opened 2016-06-23
Update README.md for updated command syntax

bgreg:
It seems that the command has changed from `bundle audit` to `bundle-audit`.  Feel free to close if I'm mistaken. 

-Greg


postmodern:
See #149. This will be in the next upcoming minor version. I temporarily reverted the README changes, since user's often refer to the README on master, instead of the [gem's documentation](http://www.rubydoc.info/gems/bundler-audit/frames).

PR #167, opened 2016-10-27
use SPDX-compatible license name

korobkov:
…to allow automatic license checks.
See https://spdx.org/licenses/

[skip ci]


korobkov:
And please, release new gem version for this change to have effect…

PR #171, opened 2017-01-02
Allow Gemfile path to be specified

rpbaptist:
This allows for bundler-audit to be run from a different directory.

I have a use case where I would like to run `bundler-audit` on several projects. Allowing the path of the project folder to be specified makes this easier without having to change directories.


leklund:
:+1: I actually created a gem (https://github.com/leklund/bauditor) to run against several projects at once. This change would make it much easier!

gurix:
I rather like to have the possibility to define the Gemfile via `BUNDLE_GEMFILE=....`.

postmodern:
bundler-audit 0.8.0 will/does support (the 0.8.0 is ready to be merged/released) support a `--gemfile-lock` option and accepts an optional `DIR` argument for specifying the project root directory.

PR #143, opened 2016-04-02
Add support for scan gems.locked

JuanitoFatas:
This Pull Request adds support for scanning new lockfile (`gems.locked`), which Bundler [introduced in 1.8.0.pre](https://github.com/bundler/bundler/blob/2-0-dev/CHANGELOG.md#180pre-2015-01-26) from [bundler/bundler@0823e10](https://github.com/bundler/bundler/commit/0823e10).

PR #184, opened 2017-12-17
Support gems.locked

kmcphillips:
Closes #143 

@JuanitoFatas @postmodern 

This takes a second try at adding `gems.locked` support.

The first commit refactors the `Scanner` class to just take the contents of a lockfile. it shouldn't have to care where it comes from.

The second commit adds a `LockfileLoader` which understands the folder, Bundler versions, and which lockfile to load.

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -16,6 +16,7 @@
 #
 
 require 'bundler/audit/scanner'
+require 'bundler/audit/lockfile_loader'
 require 'bundler/audit/version'
 
 require 'thor'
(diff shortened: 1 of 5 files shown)


kmcphillips:
Ping @JuanitoFatas @postmodern. Is this not maintained?

postmodern:
I'm leaning towards refactoring `Scanner#initialize` to accept only scanning options (ex: `:ignore`) and have `#scan` accept the lockfile. Something else would then handle the logic for find/parsing the lockfile. This would probably also mean a 1.0.0 version bump.

kmcphillips:
Oh, I think that's a much nicer API

PR #188, opened 2018-02-18
[CI] Test against Ruby 2.4 and 2.5

nicolasleger:


postmodern:
We've since switched to GitHub Actions and include 2.4 and 2.5; although 2.4 is now no longer maintained.

PR #189, opened 2018-03-02
Allow for ignoring insecure sources.

dobs:
Ignoring internal sources is already supported with a fixed IP
whitelist, but this doesn't support cases where an internal source
doesn't fall within those IPs blocks. This change allows specific
hostnames to be ignored.

Example usage:

```
bundle audit --ignore "git://private-git-host.internal/my-gem.git"
```

This change co-opts `--ignore` for simplicity, but not sure if it's better to give it its own option.


pznamensky:
Any news on this? The most desired feature at the moment.
@postmodern, probably you can take a look.

postmodern:
Just curious how `private-git-host.internal` is slipping by `#internal_source?` which resolves the IP and automatically ignores sources with private IP addresses.

pznamensky:
In our case, our private server has a global IP address, but it is closed from the Internet with firewall and access allowed only inside the same network.
That's why bundler-audit doesn't ignore this server and raise the error.

postmodern:
Hmm, I don't feel comfortable adding this feature because of your specific network configuration. I would recommend doing The Right Thing (tm) and assigning that server a non-routable subnet IP address.

pznamensky:
I don't want to argue. But The Right Thing looks a little bit subjective in this case.
You're right it's common practice to use non-routable subnets for internal servers.
But it works well in IPv4 world.
In IPv6 world things changes and I know others quite large companies except us with global IPv6 addresses assigned to internal servers. It's convenient to use global IPs and protect resources with a firewall.
Anyway, the fact that bundler-audit takes a decision about our network security and doesn't allow to correct this behavior looks strange.
This PR doesn't allow insecure sources with global IP addresses by default. So it is still safe for users. But for users with IPv6 addresses where it is common practice to use global IP address, it could be a chance to help bundler-audit to work more accurately and avoid false-positive results.

PR #190, opened 2018-04-02
Make contact link more clear

pachun:


postmodern:
and lose my impervious `s/@/ at /` spam protection? Never!

PR #199, opened 2018-10-10
Add --no-exit-on-warn flag that exits with 0 even when vulnerable

MC-Squared:
## Issue
I would like to run bundler-audit as part of our CI builds that test every branch/PR, however I do not want vulnerabilities to cause the build to be considered a failure.
Obviously, this can be achieved in bash but Brakeman, which we also run, provides a `--no-exit-on-warn` flag for this purpose.

## Fix
Added a `--no-exit-on-warn` flag that causes the CLI to exit with 0, even if vulnerabilities are found.
Error cases, such as update failing, should be unaffected by this change.

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -34,6 +34,7 @@ module Bundler
       method_option :verbose, :type => :boolean, :aliases => '-v'
       method_option :ignore, :type => :array, :aliases => '-i'
       method_option :update, :type => :boolean, :aliases => '-u'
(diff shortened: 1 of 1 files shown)


zabolotnov87:
I think it would be more useful when 0 status will be returned in case when no solutions presents yet

ingemar:
@MC-Squared You forgot to add specs for your new feature.

postmodern:
Was going to suggest just appending ` || true` to the command to suppress any exit error codes.

PR #203, opened 2018-12-12
set bundler requirement to be < 3

colby-swandale:
Hi! I'm the current release manager for Bundler, i'm going around the ruby ecosystem and seeing if there is anything i can do to help everyone prepare for the upcoming Bundler 2 release.

I noticed that the `bundler-audit` gem is locked to version 1.X at the moment. We can safely change this requirement to be `< 3`, Bundler 2 has no breaking changes that will effect this gem.

Thanks!

### DIFF
--- a/gemspec.yml
+++ b/gemspec.yml
@@ -11,4 +11,4 @@ required_rubygems_version: ">= 1.8.0"
 
 dependencies:
   thor: ~> 0.18
-  bundler: ~> 1.2
+  bundler: < 3.0


postmodern:
Went with `>= 1.2.0, < 3`, since we had issues with < 1.2.0 way back when.

petergoldstein:
@postmodern Can you publish a 0.6.1 versioned gem with this change incorporated?  That would ensure that build pipelines, etc. pick up a gem that can work with bundler 2.0

ExMember:
Travis builds are failing because bundler 2.0 requires at least Ruby 2.3.0 and Travis is running tests with 2.1.8 and 2.24.

https://github.com/rubysec/bundler-audit/blob/fade76150372f18e9c2755fa11604b9b8f6a2c2e/.travis.yml#L3-L4

Now might be a good time to add tests for 2.4, 2.5, and 2.6 in addition to dropping Rubies older than 2.3.

radar:
@ExMember I just submitted a PR to fix that here: https://github.com/rubysec/bundler-audit/pull/204.

postmodern:
Merged and now waiting on Travis. Definitely plan on releasing 0.6.1 before 0.7.0 (which still needs more specs).

postmodern:
Hmm, after changing some gem versions in the example bundles, which changed which advisory gets printed, the regular expression in `spec/integration_spec.rb` stopped matching.

https://travis-ci.org/rubysec/bundler-audit/jobs/475578889
https://github.com/rubysec/bundler-audit/commit/8db8545169195e5ddd539562894ef2a560fe42b7

postmodern:
@eric-pigeon fixed the regexp issue! I will be traveling tomorrow and the next day, so probably expect 0.6.1 to be released Thursday.

davidstosik:
@postmodern I don't want to sound pressing, but is there anything preventing the 0.6.1 release to happen?

postmodern:
I've been offline due to a really bad head cold. I prepped a release, just have to double check it, and push it up.

davidstosik:
Thank you!

PR #205, opened 2019-01-08
Fix url regular expression / update to correct gemspec operator

eric-pigeon:


postmodern:
Oh man, I was missing the `!` character. Good catch.

PR #207, opened 2019-01-17
Add support for --gemfile command line option

jnraine:
This allows bundler-audit to run files other than Gemfile.lock by passing the name of the Gemfile:

```bash
# Instead of auditing Gemfile.lock, it would audit Gemfile.next.lock
bundle-audit --gemfile Gemfile.next
```

This is useful for gems that run multiple Gemfiles for compatibility reasons or apps that use a second Gemfile to dual-boot Rails. (The latter is the case that matters to me :smile:)

Is this a change you would be open to accepting? If so, I can add tests and polish the implementation.

PR #212, opened 2019-05-04
cli: allow --file option to target a different locked gemfile

paulrbr-fl:
This changes allows to specify a file target for the audit. Instead of
the default "Gemfile.lock" file a custom gem lock file can be
specified with the new `--file gems.locked` CLI option

Replaces #207


postmodern:
We've added a `--gemfile-lock` option in the [0.8.0 branch](https://github.com/rubysec/bundler-audit/tree/0.8.0) (soon to be merged/released).

PR #122, opened 2015-11-13
Add support for an ignore file

rickypai:
Hi rubysec folks, first of all, thanks you for the great work for this gem.

I want to add the option to support an ignore file similar to `.gitignore` like `.bundlerauditignore` (name is debatable). This way, we can more easily manage which advisories to ignore in CI systems.

Thank you,

PR #215, opened 2019-07-08
Ignore file support

GUI:
This largely builds upon @rickypai's work in #122, but includes a couple small updates to get tests passing with newer versions of Bundler, and improves comment parsing a bit. In short, this adds support for a `.bundlerauditignore` file that can be used to ignore certain vulnerabilities that may not affect your application or have been explicitly handled in some other way.

Since there was never a firm answer on #122, I thought I'd open this pull request again, since we've continued to find @rickypai's fork useful. In light of https://github.com/rubysec/ruby-advisory-db/pull/390#issuecomment-509084743, it also seems like other people in that thread would find this kind functionality useful (@mmrwoods has a nice workaround in https://github.com/rubysec/ruby-advisory-db/pull/390#issuecomment-509186921, but something more official might be useful).

While the existing `--ignore` CLI flag also works for this purpose, I think having a file-based approach could provide a simpler way when running across different CI systems, and help provide more context by supporting comments in the file (so you can better document and explain why each individual advisory might be ignored).

If a YAML-based ignore file would be preferable, or there's any feedback on the file name or anything else, I'm happy to help work on this if it could help get it included in bundler-audit. Thanks!


rickypai:
thanks for building on top of the work

natebird:
I would like to see this supported as well. Thanks!

seban:
Anyone is taking care of this? Looks like this is really desired change in bundler audit.

postmodern:
In the 0.8.0 branch (soon to be merged/released) we went with another PR that added support for a unified `.bundler-audit.yml` file, with an `ignore:` Hash key.

PR #217, opened 2019-07-17
Bundle audit extended to report GitHub Security Advisories (GHSAs)

rschultheis:
## Overview

This PR extends `bundle audit` so it will report GitHub Security Advisories, in addition to the CVEs and OSVDB advisories it already reports.

I did some very mild refactoring, adding a `has_supported_identifier?` method to contain some stray logic.  I considered doing more extensive refactoring,  but in the end decided to keep the change minimal for now.

## Related links

- This PR adds the first GHSA into rubysec/advisory-db: https://github.com/rubysec/ruby-advisory-db/pull/397
- And some discussion noting the need to also extend bundler-audit, as is done here: https://github.com/rubysec/ruby-advisory-db/pull/397#issuecomment-507946376

## Sample output

This is what a GHSA report looks like:
```
$  bundle audit
Name: yard
Version: 0.9.19
Advisory: GHSA-xfhh-rx56-rxcr
Criticality: Unknown
URL: https://github.com/lsegal/yard/security/advisories/GHSA-xfhh-rx56-rxcr
Title: Possible arbitrary path traversal and file access via `yard server`
Solution: upgrade to >= 0.9.20

Vulnerabilities found!
```

CC @reedloden @jychen7 @phillmv @mveytsman

### DIFF
--- a/lib/bundler/audit/advisory.rb
+++ b/lib/bundler/audit/advisory.rb
@@ -28,6 +28,7 @@ module Bundler
                                 :cvss_v2,
                                 :cve,
                                 :osvdb,
+                                :ghsa,
                                 :unaffected_versions,
                                 :patched_versions)
 
@@ -65,6 +66,7 @@ module Bundler
           data['cvss_v2'],
           data['cve'],
           data['osvdb'],
+          data['ghsa'],
           parse_versions[data['unaffected_versions']],
           parse_versions[data['patched_versions']]
         )
@@ -86,6 +88,15 @@ module Bundler
       #
       def osvdb_id
         "OSVDB-#{osvdb}" if osvdb
+      end
+
+      #
+      # The GHSA (GitHub Security Advisory) identifier
+      #
+      # @return [String, nil]
+      #
+      def ghsa_id
+        "GHSA-#{ghsa}" if ghsa
       end
 
       #
--- a/lib/bundler/audit/advisory.rb
+++ b/lib/bundler/audit/advisory.rb
@@ -99,10 +99,6 @@ module Bundler
         "GHSA-#{ghsa}" if ghsa
       end
 
-      def has_supported_identifier?
-        cve || osvdb || ghsa
-      end
-
       #
       # Determines how critical the vulnerability is.
       #
--- a/lib/bundler/audit/advisory.rb
+++ b/lib/bundler/audit/advisory.rb
@@ -100,6 +100,16 @@ module Bundler
       end
 
       #
+      # Return a compacted list of all ids
+      def identifiers
+        [
+          cve_id,
+          osvdb_id,
+          ghsa_id
+        ].compact
+      end
+
+      #
       # Determines how critical the vulnerability is.
       #
       # @return [:low, :medium, :high]
--- a/lib/bundler/audit/advisory.rb
+++ b/lib/bundler/audit/advisory.rb
@@ -99,6 +99,10 @@ module Bundler
         "GHSA-#{ghsa}" if ghsa
       end
 
+      def has_supported_identifier?
+        cve || osvdb || ghsa
+      end
+
       #
       # Determines how critical the vulnerability is.
       #
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -109,9 +109,11 @@ module Bundler
         say "Advisory: ", :red
 
         if advisory.cve
-          say "CVE-#{advisory.cve}"
+          say advisory.cve_id
         elsif advisory.osvdb
           say advisory.osvdb
+        elsif advisory.ghsa
+          say advisory.ghsa_id
         end
 
         say "Criticality: ", :red
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -111,7 +111,7 @@ module Bundler
         if advisory.cve
           say advisory.cve_id
         elsif advisory.osvdb
-          say advisory.osvdb
+          say advisory.osvdb_id
         elsif advisory.ghsa
           say advisory.ghsa_id
         end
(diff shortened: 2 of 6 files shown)


lib/bundler/audit/scanner.rb
```diff
@@ -148,8 +148,7 @@ def scan_specs(options={})
 
         @lockfile.specs.each do |gem|
           @database.check_gem(gem) do |advisory|
-            unless (ignore.include?(advisory.cve_id) ||
-                    ignore.include?(advisory.osvdb_id))
+            if advisory.has_supported_identifier?
```
jychen7: seems this change break the `ignore` feature?
e.g. `bundle audit check --ignore CVE-2019-12345`, it should ignore `CVE-2019-12345`, but now it still `yield UnpatchedGem.new(gem,advisory)`

https://github.com/rubysec/bundler-audit/blob/6b0cfbb445c8f08b6cf592de21f3fb62a36e225a/spec/advisory_spec.rb#L97-L106

jychen7: CI pass because the first test case actually not a good one
`subject=[]` and `[].all?` returns `true` for any condition.
```
expect(subject.all? { |result| 
       result.advisory.vulnerable?(result.gem.version) 
     }).to be_truthy
```

a good test case could be
```
it "should match unpatched gems to their advisories" do
      ids = subject.map { |result| result.advisory.id }
      expect(ids).to include('OSVDB-89026')
end
```

https://github.com/rubysec/bundler-audit/blob/c3fe695f38d98138d306ef31df1decd2adc819b3/spec/scanner_spec.rb#L26-L48

lib/bundler/audit/advisory.rb
```diff
@@ -88,6 +90,19 @@ def osvdb_id
...
+      #
+      # The GHSA (GitHub Security Advisory) identifier
+      #
+      # @return [String, nil]
+      #
+      def ghsa_id
+        "GHSA-#{ghsa}" if ghsa
+      end
+
+      def has_supported_identifier?
```
reedloden: Add comment describing this

lib/bundler/audit/cli.rb
```diff
@@ -109,9 +109,11 @@ def print_advisory(gem, advisory)
         say "Advisory: ", :red
 
         if advisory.cve
-          say "CVE-#{advisory.cve}"
+          say advisory.cve_id
         elsif advisory.osvdb
           say advisory.osvdb
```
reedloden: Incorporate #126 into this and change this to `advisory.osvdb_id`

spec/advisory_spec.rb
```diff
@@ -83,6 +83,28 @@
     end
   end
 
+  describe "#has_supported_identifier?" do
```
reedloden: Should match the order as defined in the source file

lib/bundler/audit/scanner.rb
reedloden: What are you trying to do here?

rschultheis: I was misunderstanding what the `ignore` set was doing.  Basically I was setting up a scheme where the tool controls what advisory types are allowed, rather than just going off what data was present.  I thought that was what the ignore was doing, however I was quite wrong about that!  All removed now

spec/scanner_spec.rb
```diff
@@ -31,18 +31,19 @@
     subject { scanner.scan.to_a }
 
     it "should match unpatched gems to their advisories" do
+      ids = subject.map { |result| result.advisory.id }
+      expect(ids).to include('OSVDB-89025')
```
rschultheis: So this test was broken in two ways:
- `OSVDB-89026` was the advisory ID that was being used, and that is on `actionpack` and does not apply to the `unpatched_gems` Gemfiles.  It looks some past refactoring trimmed that Gemfile down significantly.
- That problem was not causing a test failure though because as @jychen7 pointed out above, the assertions in the test were not checking for the expected advisory ID, which in the later test is ignored.

Fixed this by:
- Changing the Gemfile so it uses an earlier `activerecord 3.2.10`.
- looking for OSVDB-89025 which does affect `activerecord 3.2.10`.
- also changing the ignore test to NOT expect OSVDB-89025.

Altogether this seemed like the minimal change here.

jychen7:
LGTM

reedloden:
@postmodern got time to review this?

rschultheis:
Hi @reedloden !  My apologies for dropping off this here.  Is this good to merge?  I am happy to merge it, but not sure if you or @postmodern  might prefer to?  Will you be able to take care of publishing this to RubyGems once it does merge?  This being my first PR in this repo I am not too familiar with the release process.

Thanks!

postmodern:
Forgot to submit my review... Just a minor code duplication I'd like to push down into advisory. I fully expect having to support additional Advisory ID formats in the future.

lib/bundler/audit/scanner.rb
```diff
@@ -148,9 +148,12 @@ def scan_specs(options={})
 
         @lockfile.specs.each do |gem|
           @database.check_gem(gem) do |advisory|
-            if advisory.has_supported_identifier?
-              yield UnpatchedGem.new(gem,advisory)
-            end
+            is_ignored = ignore.include?(advisory.cve_id) ||
+                         ignore.include?(advisory.osvdb_id) ||
+                         ignore.include?(advisory.ghsa_id)
+            next if is_ignored
```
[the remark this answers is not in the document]
rschultheis: cool!  I've added an `#identifiers` method, though I would be happy to change it to `#ids` if you prefer that.  Seems to better distinguish itself since there is already a `#id` attribute.

Now the check is based on that identifiers method.  Let me know if anything looks needs changing!

PR #224, opened 2019-08-29
 Allow passing Gemfile.lock paths to bundler audit

### LINKED ISSUE #178: allow passing Gemfile.lock paths to bundle audit
either as a single file or as multiple files in the same command.

### LINKED ISSUE #181: Refactor Bundler::Audit::Scanner#initialize
Looking at `#initialize` now, it seems that it should accept the scanner options (`:ignore`) and store those as instance variables. The `#scan` method would then accept the path to the `Gemfile.lock`. This would allow the Scanner object to be configured once and ran across multiple `Gemfile.lock` files.

fatkodima:
```bash
# checks Gemfile.lock by default, as now
$ bundler audit

# checks multiple lockfiles
$ bundler audit check Gemfile.lock GemfileNext.lock
```

Closes #178 
Closes #181


postmodern:
Auditing multiple `Gemfile.lock` files in one run is an interesting idea. Not sure whether users with multiple `Gemfile.lock` files would prefer to audit them individually via some set of CI tasks, or in one run; with annotated output indicated which `Gemfile.lock` file contains the vulnerable gems?

PR #225, opened 2019-08-29
Simplify .gemspec file

fatkodima:


postmodern:
The `.gemspec` file is meant to load `gemspec.yml`.

fatkodima:
@postmodern Didn't get it. Why yaml file is needed and proposed version is not working in this case?

AlexWayfer:
> The `.gemspec` file is meant to load `gemspec.yml`.

Please, describe, why it's better.

I see in this PR `+27 −72` lines of code. And a less code is the better code.

postmodern:
It separates the human readable gemspec metadata from the boilerplate gemspec code (ex: `git ls-files`). The boilerplate gemspec file is already 67 lines, so shortening it down even further feels like diminishing returns.

PR #226, opened 2019-09-25
Allow db path to be customizable

gavingmiller:
As it stands, the database must be downloaded to 

```
USER_PATH = File.expand_path(File.join(ENV['HOME'],'.local','share','ruby-advisory-db'))
```

However this breaks one of the principles of CI that everything should be installed into the local workspace directory.

This PR adds the option to optionally specify a `db_path` so that a customized directory can be set for the update.

I suspect there are other updates that may be required. Looking for feedback on whether this concept is the right way to fix this issue. 😺

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -34,6 +34,7 @@ module Bundler
       method_option :verbose, :type => :boolean, :aliases => '-v'
       method_option :ignore, :type => :array, :aliases => '-i'
       method_option :update, :type => :boolean, :aliases => '-u'
+      method_option :db_path, :type => :string, :aliases => '-p', default: nil
 
       def check
         update if options[:update]
@@ -66,7 +67,7 @@ module Bundler
       def update
         say("Updating ruby-advisory-db ...") unless options.quiet?
 
-        case Database.update!(quiet: options.quiet?)
+        case Database.update!(quiet: options.quiet?, db_path: options.db_path)
         when true
           say("Updated ruby-advisory-db", :green) unless options.quiet?
         when false
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -34,7 +34,7 @@ module Bundler
       method_option :verbose, :type => :boolean, :aliases => '-v'
       method_option :ignore, :type => :array, :aliases => '-i'
       method_option :update, :type => :boolean, :aliases => '-u'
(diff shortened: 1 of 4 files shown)


lib/bundler/audit/cli.rb
```diff
@@ -34,6 +34,7 @@ class CLI < ::Thor
       method_option :verbose, :type => :boolean, :aliases => '-v'
       method_option :ignore, :type => :array, :aliases => '-i'
       method_option :update, :type => :boolean, :aliases => '-u'
+      method_option :db_path, :type => :string, :aliases => '-p', default: nil
```
pxlpnk: Having the proper default here vs `nil` is a nicer way, it avoids passing around nil.

mokhan: @pxlpnk are you suggesting? 

```ruby
default: File.join(Dir.home, ".local", "share", "ruby-advisory-db")
```

```irb
irb(main):001:0> File.join(Dir.home, '.local', 'share', 'ruby-advisory-db')
=> "/home/mokha/.local/share/ruby-advisory-db"
```

pxlpnk: I think that could work.  Or use a variable and assign it that value.

gavingmiller: @mokhan @pxlpnk updates pushed.

lib/bundler/audit/cli.rb
```diff
@@ -34,6 +34,7 @@ class CLI < ::Thor
       method_option :verbose, :type => :boolean, :aliases => '-v'
       method_option :ignore, :type => :array, :aliases => '-i'
       method_option :update, :type => :boolean, :aliases => '-u'
+      method_option :db_path, :type => :string, :aliases => '-p', :default => File.join(Dir.home, '.local', 'share', 'ruby-advisory-db')
```
mokhan: I didn't notice this earlier but it looks like you can use `Bundler::Audit::Database::USER_PATH`.

gavingmiller:
> integration tests are failing due to new advisories

Interestingly enough the `rails-html-sanitizer` failures are what led to this PR. I believe those failures are actually because of an outdated database being picked up. I may have misattributed the issue, but we were seeing this on our CI too, but not locally when running tests.

postmodern:
I ended up adding the `BUNDLER_AUDIT_DB` env variable in the 0.7.0 branch (9375524d845d3444a7175857c6e15436a7f41c18) to allow globally changing the DB path. I also need change `Scanner#initialize` to receive a database object, that way we can initialize the DB then pass it into the `Scanner.new`

PR #228, opened 2019-10-21
Allow to ignore specific Gemfile groups

### LINKED ISSUE #195: Feature request: Allow group selection
I'd like to be able to select a *subset* of gems based on the groups in the Gemfile.

In my case, [rubyzip has an unresolved vulnerability](https://github.com/rubyzip/rubyzip/pull/371) but since I only use it in test or development, it doesn't matter to me.

I'd like to be able to say "report only gems that are ALWAYS required or are in the production group".  Alternatively, being able to "ignore all gems that are ONLY used in test or development" would work too.

fatkodima:
Algorithm:
1. get all groups from Gemfile (`gemfile.groups`)
2. remove ignored groups from the above (`groups_to_check`)
3. get all gem specifications for gems in groups to check (`parent_dependencies`)
4. using breadth-first search, get the remaining gem specifications

Closes #195


fatkodima:
ping @postmodern @reedloden

postmodern:
FYI, in the [0.8.0 branch](https://github.com/rubysec/bundler-audit/tree/0.8.0) (soon to be merged/released) support for a `.bundle-audit.yml` configuration file was added with support for a `ignore:` key to ignore individual advisories by ID.

I'm hesitant about ignoring whole groups, as that could allow vulnerabilities to go un-patched/un-mitigated; attackers don't care if a vulnerable gem is only activated during development if they can still somehow exploit it. Ignoring by individual advisory, where you confirm that you've mitigated it somehow, is much safer.

Issue #175, opened 2017-05-13
actionpack 5.1.1 vulnerable. Upgrade to lower version 5.0.0.beta1.1


rdeavila:
Some `action*` projects are reporting as insecure on version `5.1.1`. But the solution is upgrade to a lower version 5.0.0.beta1.1.

This is correct?

```
Name: actionpack
Version: 5.1.1
Advisory: CVE-2016-0751
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/9oLY_FCzvoc
Title: Possible Object Leak and Denial of Service attack in Action Pack
Solution: upgrade to ~> 5.0.0.beta1.1, ~> 4.2.5.1, ~> 4.1.14.1, ~> 3.2.22.1

Name: actionpack
Version: 5.1.1
Advisory: CVE-2015-7576
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/ANv0HDHEC3k
Title: Timing attack vulnerability in basic authentication in Action Controller.
Solution: upgrade to ~> 5.0.0.beta1.1, ~> 4.2.5.1, ~> 4.1.14.1, ~> 3.2.22.1

Name: actionview
Version: 5.1.1
Advisory: CVE-2016-0752
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/335P1DcLG00
Title: Possible Information Leak Vulnerability in Action View
Solution: upgrade to ~> 5.0.0.beta1.1, ~> 4.2.5.1, ~> 4.1.14.1, ~> 3.2.22.1

Name: activemodel
Version: 5.1.1
Advisory: CVE-2016-0753
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/6jQVC1geukQ
Title: Possible Input Validation Circumvention in Active Model
Solution: upgrade to ~> 5.0.0.beta1.1, ~> 4.2.5.1, ~> 4.1.14.1

Name: activerecord
Version: 5.1.1
Advisory: CVE-2015-7577
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/cawsWcQ6c8g
Title: Nested attributes rejection proc bypass in Active Record
Solution: upgrade to ~> 5.0.0.beta1.1, ~> 4.2.5.1, ~> 4.1.14.1, ~> 3.2.22.1
```

wtfiwtz:
[...]

PR #230, opened 2019-10-22
Fail for Database.update! when git is not available

### LINKED ISSUE #158: Database.update! shall fail when git not installed
I am using the tool (thanks for providing it) within an alpine linux docker container.

I wondered why the update feature did not work. 
It failed without error.

The reason is, that `Database.update!` fails silently when git is not installed (https://github.com/rubysec/bundler-audit/blob/master/lib/bundler/audit/database.rb#L94-L104)
- `.update!` should check if `git` is available
- `.update!` should return false or an error when `git` is not available on the system

fatkodima:
Closes #158

### DIFF
--- a/spec/cli_spec.rb
+++ b/spec/cli_spec.rb
@@ -46,6 +46,33 @@ describe Bundler::Audit::CLI do
         end
 
       end
+
+      context "when git is not installed" do
+        before do
+          expect(Bundler::Audit::Database).to receive(:update!).and_return(nil)
+          expect(Bundler).to receive(:git_present?).and_return(false)
+        end
+
+        it "prints failure message" do
+          expect do
+            begin
+              subject.update
+            rescue SystemExit
+            end
+          end.to output(/Git is not installed!/).to_stdout
+        end
+
+        it "exits with error status code" do
+          expect {
+            # Capture output of `update` only to keep spec output clean.
+            # The test regarding specific output is above.
+            expect { subject.update }.to output.to_stdout
+          }.to raise_error(SystemExit) do |error|
+            expect(error.success?).to eq(false)
+            expect(error.status).to eq(1)
+          end
+        end
+      end
     end
 
     context "--quiet" do
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -73,6 +73,10 @@ module Bundler
           say "Failed updating ruby-advisory-db!", :red
           exit 1
         when nil
+          unless Bundler.git_present?
+            say "Git is not installed!", :red
+            exit 1
+          end
           say "Skipping update", :yellow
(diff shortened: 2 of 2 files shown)


pxlpnk:
Raising an exception and failing with a stack trace might be not a good way of handling this situation.
A more and better defined way would be to print an error message and end the process with an exist status `!=0`

fatkodima:
@pxlpnk Updated.
Now `cli.rb` checks that git is not installed.

Or should `Database.update!` still raise some custom error and in `cli.rb` we rescue it and print message?

pxlpnk:
I think its good the way is it now. Maybe in the future with different output formaters this would have to be changed.

pxlpnk:
@postmodern is there a way to get this merged?

and0x000:
possibly related to https://github.com/rubysec/bundler-audit/issues/175 where it is also mentioned, that updating the DB is skipped.

pxlpnk:
@reedloden How do we get contributions into bundler audit?

ZimbiX:
LGTM; I'd probably have used an exception as mentioned (whenever a call to Git is made where it's not installed), but this looks fine to me. Definitely good not to be giving people a false sense of security. Cheers for the fix

@postmodern Merging would be much appreciated =) Thanks

reedloden:
Can we split the other changes and the DB update?

fatkodima:
@reedloden Updated.

spec/bundle/insecure_sources/Gemfile
```diff
@@ -1,4 +1,4 @@
 source 'http://rubygems.org'
 
-gem 'rails'
+gem 'rails', '~> 5.2'
```
reedloden, line 3: This doesn’t seem related. Can this be removed?

spec/cli_spec.rb
```diff
@@ -46,6 +46,33 @@
         end
 
       end
+
+      context "when git is not installed" do
+        before do
+          expect(Bundler::Audit::Database).to receive(:update!).and_return(nil)
```
reedloden, line 52: The specs were all updated recently. Please see about conforming this and any other code to the recent style changes. For example, probably can use described_class here?

spec/bundle/insecure_sources/Gemfile
fatkodima: removed

spec/cli_spec.rb
fatkodima: `described_class` is `Bundler::Audit::CLI` in this case, so cannot be used
I do not see inconsistencies with surrounding code.

reedloden: Thanks! I was reviewing from my phone, so I couldn't see full file context. :-)

PR #236, opened 2019-12-18
Fix Scanner#internal_source? w/ Bundler::URI::HTTP

milgner:
Trying to use bundler-audit with Bundler 2.1.1 fails because
`Scanner#internal_source?` gets passed a `Bundler::URI::HTTP` instance.

```
uri/common.rb:745:in `URI': bad argument (expected URI object or URI string) (ArgumentError)
```

This patch uses `#to_s` on the passed object to work around that issue.

### DIFF
--- a/lib/bundler/audit/scanner.rb
+++ b/lib/bundler/audit/scanner.rb
@@ -167,7 +167,7 @@ module Bundler
       # @return [Boolean]
       #
       def internal_source?(uri)
-        uri = URI(uri)
+        uri = URI.parse(uri.to_s)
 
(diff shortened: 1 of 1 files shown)


milgner:
Looks like CI is [already failing on `master`](https://travis-ci.org/rubysec/bundler-audit/jobs/620851425)?

postmodern:
Odd, it appears Bundler 2.x has vendored the entire `URI` library in. In the future we should standardize on using `Bundler::URI`, so that we're consistent with their code.

Issue #235, opened 2019-12-16
ArgumentError: bad argument (expected URI object or URI string)


slayer:
bundler-audit starts failing today at CI: 

```
$ cd /app && bundle exec bundle-audit update && bundle exec bundle-audit
Updating ruby-advisory-db ...
Cloning into '/home/app/.local/share/ruby-advisory-db'...
Updated ruby-advisory-db
ruby-advisory-db: 422 advisories
bundler: failed to load command: bundle-audit (/app/vendor/bundle/ruby/2.6.0/bin/bundle-audit)
ArgumentError: bad argument (expected URI object or URI string)
  /usr/local/lib/ruby/2.6.0/uri/common.rb:739:in `URI'
  /app/vendor/bundle/ruby/2.6.0/gems/bundler-audit-0.6.1/lib/bundler/audit/scanner.rb:170:in `internal_source?'
  /app/vendor/bundle/ruby/2.6.0/gems/bundler-audit-0.6.1/lib/bundler/audit/scanner.rb:113:in `block (2 levels) in scan_sources'
  /app/vendor/bundle/ruby/2.6.0/gems/bundler-audit-0.6.1/lib/bundler/audit/scanner.rb:112:in `each'
  /app/vendor/bundle/ruby/2.6.0/gems/bundler-audit-0.6.1/lib/bundler/audit/scanner.rb:112:in `block in scan_sources'
  /app/vendor/bundle/ruby/2.6.0/gems/bundler-audit-0.6.1/lib/bundler/audit/scanner.rb:102:in `map'
  /app/vendor/bundle/ruby/2.6.0/gems/bundler-audit-0.6.1/lib/bundler/audit/scanner.rb:102:in `scan_sources'
  /app/vendor/bundle/ruby/2.6.0/gems/bundler-audit-0.6.1/lib/bundler/audit/scanner.rb:74:in `scan'
  /app/vendor/bundle/ruby/2.6.0/gems/bundler-audit-0.6.1/lib/bundler/audit/cli.rb:44:in `check'
  /app/vendor/bundle/ruby/2.6.0/gems/thor-0.19.4/lib/thor/command.rb:27:in `run'
  /app/vendor/bundle/ruby/2.6.0/gems/thor-0.19.4/lib/thor/invocation.rb:126:in `invoke_command'
  /app/vendor/bundle/ruby/2.6.0/gems/thor-0.19.4/lib/thor.rb:369:in `dispatch'
[...]

PR #240, opened 2020-01-07
Fix internal source check in scanner

ThomasKoppensteiner:
This change converts the input `uri` to a string within the `internal_source?` method to avoid the `ArgumentError` issue from #235.


ThomasKoppensteiner:
Build on the master are also failing. How should we proceed here?

ThomasKoppensteiner:
PR #236 did a similar fix.

PR #253, opened 2020-03-25
#218 Add the ability to use dot config file to control the ignore list

### LINKED ISSUE #218: Read ignores from e.g., a `.bundler-audit.ignore` file
GitHub issues security alerts for Ruby projects based on the `Gemfile.lock` within a given repo. I suspect they may be using `bundler-audit` but users don't have any control over the flags they pass on the command line.

It'd be great if bundler-audit could read something comparable to `.gitignore` files but with a name particular to bundler-audit itself. That way users could inform GitHub's systems when particular CVEs have been remediated in a way that bunder-audit is unable to automatically verify.

Case in point: all versions of the `omniauth` gem are vulnerable in Rails projects unless the `omniauth-rails_csrf_protection` is also included and HTTP verbs other than 'POST' are disabled (e.g., throw 404s) in the omniauth endpoints. Maybe the detection logic for `CVE-2015-9284` could be made more sophisticated but it seems like it'd also be useful if we had a way to just tell GitHub that it's been handled in case there are other CVEs where remediation is difficult to automatically determine.

jebentier:
This is addressing #218 by adding the ability to have a `.bundler-audit.yml` configuration file at the root of the project.  It is a yaml file and expects the ignore property to have an array of CVEs underneath it.  This issue seems to be desired by quite a few and would like to get this implemented so that it can start being used by the Github dependency bot when it is doing it's security scans.  

This is also quite extensible in the future if more configuration is added to the gem, it can also be reflected through the yaml file.

In order to get unit tests to pass again a few things needed to be done:
1. The advisory database needed to be updated. On master the advisory database timestamp file was updated but the sha of the database submodule was not. This was causing one of the 2 failures in `database_spec.rb`.
2. Some usage of `subject` needed to be removed from `database_spec.rb` for the dynamic `Database#path` tests.  Using `subject` is very helpful when you're making very thin and static tests.  But it should be known that the value `subject` is memoized the first time it is called.  This unfortunately means that if you expect the value returned by the `subject` block to change throughout the test, you cannot use this method for simplifying the tests.
3. Due to the update of the advisory database, the `secure` test bundle was now failing due to some new vulnerabilities, so the bundle was updated.

Closes #218


bencrouse:
Hey, this would be handy for us, any updates on when this might get released?

dustym:
Agreed, this would be pretty nice on days like this when there are a bunch of CVEs dropped.

spec/database_spec.rb
```diff
@@ -8,29 +8,27 @@
...
   describe "path" do
-    subject { described_class.path }
-
     it "it should be a directory" do
-      expect(File.directory?(subject)).to be_truthy
+      expect(described_class.path).to be_truthy
     end
 
-    it "should prefer the user repo, iff it's as up to date, or more up to date than the vendored one" do
+    it "should prefer the user repo, iff it's as up to date, or more up to date than the vendored one", focus: true do
```
mauro-oto, line 15: This should probably not have `focus: true`.

jebentier: Thank you for catching that, I've removed it.

reedloden:
@jebentier can you rebase this? I think all your unit test failures should have been addressed now.

postmodern:
@jebentier please create a `Configuration`/`Config` class to handle loading/validating/using the config file data. Also since this is a new feature, this should go into 0.7.0 not master.

postmodern:
Also this should be targeted to the 0.8.0 branch, since we've just released 0.7.0.

jebentier:
@reedloden @postmodern sorry for the delay on this. I've gone ahead and rebased off of `0.8.0` and made the changes requested for extracting out `Bundler::Audit::Configuration` and adding validation for the ignore configuration.

postmodern:
@jebentier also sorry for the delay and more change request.

1. All new classes and methods need YARDocs.
2. There's an easier way to safely parse and validate YAML. Use `YAML.parse` instead of `YAML.load`. See the [YAML.parse](https://github.com/rubysec/bundler-audit/blob/YAML.parse/lib/bundler/audit/advisory.rb#L43-L163) branch (which I should probably merge as well) on how to explicitly check/convert YAML nodes to Ruby primitives.

jebentier:
@postmodern No problem at all, happy to take care of any requests necessary for the community to feel comfortable getting this change into the project.  I've gone ahead and added YARDocs to the `Configuration` class, and converted to using `YAML.parse` with additional validation instead of `YAML.load` in the `Configuration.from_yaml` composable method.  Please let me know if there is anything else I can do.

fabn:
@postmodern Any ETA for release 0.8? Thanks.

jeffxor:
Any chance we could add some documentation to this feature for the format of the file?

postmodern:
@fabn Recently setup GitHub Actions on `master` branch and currently trying to get specs passing. Then I will merge 0.8.0.

@jeffxor If anyone wants to take a stab at documenting the new configuration file in either the README or a separate file, that would be one less thing for me to do. See [this](https://github.com/rubysec/ruby-advisory-db#schema) or [this](https://github.com/ruby-ore/ore/blob/master/GemspecYML.md) as examples of how to document YAML schemes.

postmodern:
Got the failing specs down to CLI specs, which are failing to load thor (probably yet another bundler issue). Making progress...

<pre>
  5) CLI update when advisories update successfully should print status
     Failure/Error: raise "FAILED #{command}\n#{result}" if $?.success? == !!options[:fail]
     
     RuntimeError:
       FAILED /data/home/postmodern/code/rubysec/bundler-audit/bin/bundler-audit update
       /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:83:in `require': cannot load such file -- thor (LoadError)
</pre>

postmodern:
Got specs passing on 0.8.0 by removing `Bundler.with_clean_env` from the `sh` helper in `spec_helper.rb`. Now writing specs for `Bundler::Audit::CLI::Formats::Text#print_report` and `Bundler::Audit::CLI::Formats::JSON#print_report`. This will help reduce the pressure on `spec/integration_spec.rb` and allow us to more cleanly test the output formatting.

postmodern:
Specs written. Will merge the 0.8.0 branch, then get GitHub Actions CI passing, then we can do a pre-release to let people test the new version before a 0.8.0 release.

postmodern:
One issue I just spotted. Users in the past have requested the ability to ignore insecure source URIs. In response I added logic to just filter out any source URIs that pointed to internal/local IP subnets, like those from VPNs or LANs. The configuration file schema only supports an `ignore: Array[String]` key. If we were to support ignoring source URIs, then we'd need to split `ignore:` into two sub-keys:

    ignore:
      advisories:
        - ...
      sources:
        - ...

Or, we could simply draw the line about what should or shouldn't be configurable, and agree that unencrypted source URIs should not be ignored; unless of course they point to a local/internal IP address like with a VPN or LAN.

postmodern:
Added really basic example/documentation for the `.bundler-audit.yml` config file:
https://github.com/rubysec/bundler-audit/blob/0.8.0/README.md#configuration-file

jebentier:
I think expanding the `ignore` structure will be powerful. Something in addition to that which we've already been doing at Invoca is doing temporary ignoring of CVEs so that we can push off the alert to a later date.  This change would be simple to add with the new YAML configuration file using the structure:

```yaml
ignore:
  - CVE-123-1234 # this is ignored forever
  - id: CVE-456-4567 # this is only ignored until the first of the year 2021
    until: 2021-01-01
```

Or with your other additions you mentioned above it would go under the `ignore.advisories` section.  Previously we were adding this functionality with a gem wrapper, but now that this is released in an RC, I'm looking at what it would take to get it implemented. See any interest in getting this `ignore until` functionality added here?

postmodern:
@jebentier could you give me a security explanation as to why you'd want to ignore an advisories up until a certain date? Unless the vulnerability is patched or mitigated, you are still vulnerable. It's not like attackers will politely wait until January 1st, 2021 to attack you. If it's more of a way to set a reminder for yourself to replace any mitigations/workarounds with a more permanent solution (upgrading a gem or removing code), could one of those code reminder/"time bomb" tools/libraries be used instead?

jebentier:
@postmodern sorry for the delayed response here. We put CVE audits as a blocker to deploys to make sure that new CVEs are not introduced without knowing.  With doing this, we've both increased visibility and decreased the time for applying the necessary patches. There are times however that a new CVE is reported (especially when we're using large libraries like rails) that is not currently exposed by the way we use the library. We do not want to prevent deploys due to the risk assessment we've made, but we also don't want to forget about the CVE and miss the patch entirely incase the vulnerability were to be exposed by future work.  So we essentially put it into our technical debt backlog with a deadline, and put that deadline into the temporary ignoring of the CVE. This builds out the reminder to us that we need to reassess if necessary and get the CVE patched. Using another system for the reminder is definitely possible, but using multiple disjoint systems will always be less effective and more error prone than using a single system, this is why we've built the wrapper of this gem to combine it with our CI pipeline.

postmodern:
@jebentier ah that makes more sense. I still think there's probably a better way to set a reminder about upgrading a dependency due to non-impacting advisories (Calendar? Ticket System?) and that bundler-audit managing time-based triggers seems beyond-scope. It probably would be a good practice to put a comment in the YAML file explaining why the `ignore` entry was added. Perhaps maybe even a `reason: ` key could be added?

    - id: CVE-YYYY-XXXX
      reason: We do not use the vulnerable module, yada yada ...

jebentier:
@postmodern Expanding the array to be a hash to hold additional information about the ignore does sound like a good idea, and should be done before the initial release to avoid complex or breaking changes with the file format.  Reason sounds like a good addition, additionally it might be helpful to be reminded by the CLI output what the skipped CVEs were with the `Skipped because #{reason}`.

postmodern:
Should bundler-audit support both `Array<String>` and `Array<Hash>` styles of `ignore:` YAML? `- id: CVE-....` seems like a lot to type if a user just wants to ignore a CVE; `- CVE ...` is more succinct. Or should bundle-audit require `id:` and `reason:` Hash keys for each entry and only support parsing `Array<Hash>`?

### Only Array<String>

```yaml
ignore:
  - CVE-...
  - CVE-...
```
  
 ### Mixed
  
  ``` yaml
ignore:
    - CVE-...
    - id: CVE-...
      reason: ...
      ...
```

### Only Array<Hash>

```yaml
ignore:
  - id: CVE-...
    reason: ...
  - id: CVE-...
    reason: ...
```

jebentier:
I think it would be good to support both, although it is simpler (code wise) to just support `id` as a required Hash key under the `ignore` array. It might be worth while to think about any other fields that would be useful to keep track of in the ignoring of CVEs. Reason is definitely a nice one, but I don't think it should be required.  YAML also supports comments, so the reason can be placed in the config file through the use of comments.

PR #258, opened 2020-05-15
Use non-zero exit code if there's an argument error when running through CLI

mauro-oto:
Our CircleCI build was passing (and failing silently) when running bundle-audit since we were not passing the right arguments to the `start` method.

Now when we run our custom bundler audit Rake task, which uses the CLI directly, we get a non-zero exit error when we pass the wrong arguments:

Without this PR:
```
➜  bundle exec rake bundle:audit
ERROR: "rake check" was called with arguments ["--ignore CVE-2015-9284"]
Usage: "rake check"
➜ echo $?
0
```

With this PR:
```
➜  bundle exec rake bundle:audit
ERROR: "rake check" was called with arguments ["--ignore CVE-2015-9284"]
Usage: "rake check"
➜ echo $?
127
```

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -156,6 +156,10 @@ module Bundler
         say
       end
 
+      def self.exit_on_failure?
+        true
+      end
+
     end
   end
 end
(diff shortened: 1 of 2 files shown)


mauro-oto:
@reedloden any chance you can review/merge this PR? Thank you!

mauro-oto:
cc @postmodern

PR #264, opened 2020-06-10
Docs: Install with bundler

jaredbeck:
My team requires all gems to be listed in `Gemfile`. Also, we are installing from source right now so that we can use the latest `thor` (pending the release of `0.7`).

Maybe we're the only ones, but just in case it helps someone else ..

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,19 @@ task default: 'bundle:audit'
 
     $ [sudo] gem install bundler-audit
 
+### Install using bundler
+
+Installation with `gem` (above) is preferred.
+
+```ruby
+# Gemfile
+group :development do
+  gem 'bundler-audit', require: false
+  # or, from source:
+  gem 'bundler-audit', github: 'rubysec/bundler-audit', require: false
+end
+```
+
 ## Contributing
 
 1. Clone the repo


README.md
```diff
@@ -136,6 +136,40 @@ task default: 'bundle:audit'
...
+
+It is possible, but after installation, you must init the 
+database submodule.
+
+```ruby
+# Gemfile
+group :development do
+  gem 'bundler-audit',
+    github: 'rubysec/bundler-audit',
+    ref: 'abcd1234', # for security, always specify ref
```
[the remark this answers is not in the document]
jaredbeck: > It's a problem of these workers and their coworkers.

I mostly agree, and I will remove the `ref`.

> .. You can leave only :github: and :require ..

Do you want me to remove the `group :development`? I'm worried that if I remove it, beginners may make the mistake of including it in their production builds. I think it's a common mistake.

> Instructions about git submodules are informative (if true, I didn't check) ..

Yeah, I haven't tested that part thoroughly either. I was getting an from git when running eg. `bundle update`, 

```
error: could not lock config file .git/modules/data/ruby-advisory-db/config: No such file or directory
```

.. and the `submodule update` seems to have fixed that error, but that's all I know. `bundle-audit check --update` produces the expected `git` output, indicating that it is updating `ruby-advisory-db`, so all *seems* to be working.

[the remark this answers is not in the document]
AlexWayfer: > Do you want me to remove the `group :development`?

No, I'm about `gem ...` options.



> I was getting an from git when running eg. `bundle update`

Maybe you didn't run `bundle-audit check --update` after `bundle install` and `bundle update`? I'm not sure, I can try to check it later (today I hope).

jaredbeck:
Rebased, and `ref:` option removed

postmodern:
bundler-audit 0.7.0 (and subsequent 0.7.0.1) have finally been released with the thor dependency fix.

README.md
```diff
@@ -136,6 +136,37 @@ task default: 'bundle:audit'
...
+  gem 'bundler-audit', github: 'rubysec/bundler-audit', require: false
+end
+```
+
+```bash
+bundle install
+cd `bundle show bundler-audit`
+git submodule update --init
+cd -
+```
```
AlexWayfer, line 168: I can't reproduce any problem for these steps. Can you please describe and provide some instructions?

jaredbeck: Which step do you suspect is unnecessary? The `submodule update`? I recall that fixed an error saying that it couldn't find the "database" (I don't remember the name of the submodule) But, if it's working for you without the `submodule update` then let's drop that part?

AlexWayfer: > Which step do you suspect is unnecessary?

I wrote "for these steps", all steps from this code block. Except the first one, of course.

So, I've deleted all installed `bundler-audit` copies (I hope), ran `bundle install` and `bundle-audit check --update` works.

jaredbeck: Cool, thanks for testing that. Want me to delete the suspected unnecessary commands?

AlexWayfer: > Cool, thanks for testing that. Want me to delete the suspected unnecessary commands?

I think, yes. I see no need for them. If somebody experiences problems or will to — we can investigate and solve their problems, probably with such lines for installation, but maybe not (in another way, depending on a problem).

README.md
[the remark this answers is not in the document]
AlexWayfer: Now I have the same warnings:

```
> bundle-audit check --update
Updating ruby-advisory-db ...
From https://github.com/rubysec/ruby-advisory-db
 * branch            master     -> FETCH_HEAD
Already up to date.
Updated ruby-advisory-db
ruby-advisory-db: 466 advisories
No vulnerabilities found

> bundle update
Fetching https://github.com/rubysec/bundler-audit.git
error: could not lock config file .git/modules/data/ruby-advisory-db/config: No such file or directory
warning: Could not unset core.worktree setting in submodule 'data/ruby-advisory-db'
Fetching gem metadata from https://rubygems.org/.........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
...
```

`git submodule update` fixes. This is strange.

postmodern:
Note that in bundler-audit 0.8.0 (see the 0.8.0 branch) we removed the vendored ruby-advisory-db, in favor of auto-downloading/auto-updating it, so no issues with `git submodule`. Simply add `gem 'bundler-audit'` to the `Gemfile`.

PR #268, opened 2020-06-22
support for ruby 2.4 has ended

mariuz:
https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/


postmodern:
Note I have since switched to GitHub Actions and removed `.travis.yml`.

Issue #267, opened 2020-06-14
Git is not installed!


Domon:
Hi,

`bundle-audit check --update` failed to run in our build process suddenly and returned the following error message.

> Git is not installed!

I checked the Changelog but couldn't find any related information or upgrading suggestion.
https://github.com/rubysec/bundler-audit/blob/e64c5c7197edcb5b19bb94f63a39b6819a26e03b/ChangeLog.md

I wonder if that error comes from a PR merged recently.
https://github.com/rubysec/bundler-audit/pull/230/files#diff-f3ec31a31911b9265366828e72a41e68R77

I'm curious why we were able to run `bundler-audit` without coming across this error message previously.

Does `bundler-audit` require `git` to run? Is this something new? 🤔 

Thanks.

reedloden:
`bundler-audit` has always required `git` for keeping its advisory database updated. The repository comes from https://github.com/rubysec/ruby-advisory-db. Without `git`, any use of `bundler-audit` would not be using an updated database, meaning any checks run would likely miss all vulnerabilities added since the last `bundler-audit` gem release.

reedloden:
See also https://github.com/rubysec/bundler-audit/issues/259.

Domon:
Thanks @reedloden for the explanation. 🙇 

Out of curiosity, I installed `0.6.1` and tried the same command. The command was executed successfully with a message saying that the database was not updated.

```
$ bundle-audit check --update
Updating ruby-advisory-db ...
Skipping update
ruby-advisory-db: 287 advisories
No vulnerabilities found
```

In comparison, the command failed in `0.7.0.1`.

```
$ bundle-audit check --update
Updating ruby-advisory-db ...
Git is not installed!
```

I guess most people would like to run the command with the latest database. It's just not very obvious in the error message and the breaking change was not highlighted in the change log.

I wonder if the latest database can be downloaded without `git` though. 🤔

mbirman:
@Domon are you using [andrewmcodes/bundler-audit-action](https://github.com/andrewmcodes/bundler-audit-action)?

Domon:
> @Domon are you using [andrewmcodes/bundler-audit-action](https://github.com/andrewmcodes/bundler-audit-action)?

No. Good to know there is a Github Action for bundler audit though.

enthusiasmus:
For us it is the most important feature ever! No one knew that the advisories db doesn't get updated without git.

postmodern:
[...]

PR #270, opened 2020-07-13
Doc: Add git as part of the requirement

rizalmuthi:
Hei ya team, 
I got an error saying that `Git is not installed!`
And I realized that:
1. There is a current issue with this: https://github.com/rubysec/bundler-audit/issues/267
2. In the requirement, git is not listed as part of it

We also assumed that this gem does not require git to run, so now our deployment is a little bit broken with the latest version.
By adding git as part of the requirement I believe will help the other developers in the future.

Cheers


reedloden:
Thanks!

PR #273, opened 2020-09-09
Update README.md

kaskas:
Adding CodeFactor's automated CodeQuality Badge


postmodern:
Are there any selling points and why we should use CodeFactor over CodeClimate?

Issue #287, opened 2021-01-27
Fix specs due to new thor 1.1.0


postmodern:
Specs suddenly began failing after thor 1.1.0 was released. Likely thor's new internal behavior is breaking our `expect().to receive()` stubs that we use to populate CLI's `options`.

postmodern:
It also appears that Thor now disables _all_ output if `--quiet` is specified. To get around this, we'll want to use `error` to output error messages. Although, `error` doesn't seem to support ANSI color output or highlight it's output as red by default.

postmodern:
Fixed in 2eea92707af22639cde2ef677af851fc71ce6699 and 2488b18380ba66075203215182e971dbab954f03.

PR #291, opened 2021-02-13
Add Truffleruby head to CI

gogainda:


gogainda:
@postmodern test failures now consistent with other rubies

postmodern:
@gogainda see issue #287 concerning other spec failures. Has the [secure_random](https://github.com/rubysec/bundler-audit/runs/1684264130?check_suite_focus=true) bug been fixed in truffleruby-head yet?

gogainda:
@postmodern should be (in head thou)

PR #296, opened 2021-04-19
Use YAML.safe_load

ooooooo-q:
I changed `YAML.load_file` to `YAML.safe_load` because it can deserialize unexpected classes.

RCE can occur if there are no restrictions on the classes that can be deserialized, if the ruby-advisory-db git repository contains a specially crafted yaml file, or if it is crafted into a local cache.

FYI https://gist.github.com/staaldraad/89dffe369e1454eedd3306edc8a7e565

### DIFF
--- a/lib/bundler/audit/advisory.rb
+++ b/lib/bundler/audit/advisory.rb
@@ -45,7 +45,7 @@ module Bundler
       #
       def self.load(path)
         id   = File.basename(path).chomp('.yml')
-        data = YAML.load_file(path)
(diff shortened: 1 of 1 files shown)


postmodern:
Merged. There was an older branch called [YAML.parse](https://github.com/rubysec/bundler-audit/commits/YAML.parse) where we attempted to explicitly coerce each node in the parsed YAML tree. `YAML.safe_load` also works just as fine.

PR #301, opened 2021-05-13
Fix text format output 

gonzoyumo:
Since version **v0.8.0**, the default output (text) of the CLI is impacted by a regression.

Advisories with CVSS v3 scores have a broader range of Criticality values than with CVSS v2. `None` and `Critical` values are currently missing from the `print_advisory` method in [lib/bundler/audit/cli/formats/text.rb](https://github.com/rubysec/bundler-audit/blob/v0.8.0/lib/bundler/audit/cli/formats/text.rb#L79-L85).

This PR adds cvss_v3 value to the fixture and ensure to test output for both v2 and v3 values. This is a simle fix which keeps current architecture as is but it might be welcome to have a single place where Criticality values are defined and how they should be printed, to avoid such discrepancy in the future.

Thanks.

### DIFF
--- a/spec/cli/formats/text_spec.rb
+++ b/spec/cli/formats/text_spec.rb
@@ -104,39 +104,110 @@ describe Bundler::Audit::CLI::Formats::Text do
           end
         end
 
-        context "when Advisory#criticality is :low" do
-          let(:advisory) do
-            super().tap do |advisory|
-              advisory.cvss_v2 = 0.0
+        context "when CVSS v3 is present" do
+          context "when Advisory#criticality is :none (cvss_v3 only)" do
+            let(:advisory) do
+              super().tap do |advisory|
+                advisory.cvss_v3 = 0.0
+              end
+            end
+
+            it "must print 'Criticality: None'" do
+              expect(output_lines).to include("Criticality: None")
+            end
+          end
+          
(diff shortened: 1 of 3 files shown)


gonzoyumo:
@postmodern could you please have a look at this bugfix? Thanks!

spec/cli/formats/text_spec.rb
```diff
@@ -104,39 +104,110 @@
...
+            end
           end
         end
 
-        context "when Advisory#criticality is :high" do
+        context "when CVSS v2 is present" do
           let(:advisory) do
             super().tap do |advisory|
-              advisory.cvss_v2 = 10.0
+              advisory.cvss_v3 = nil
```
postmodern, line 173: Is this test testing CVSSv2 or CVSSv3?

gonzoyumo:
Thanks @postmodern for merging this fix. Is there any planned release of a `0.8.1` version?

postmodern:
I ended up making a few minor additions to the rake task, so possibly a 0.9.0 is on the way? I could possibly make a 0.8.1 branch and cherry-pick in only the patch commits.

gonzoyumo:
@postmodern thanks! Whatever works best for you.

Issue #308, opened 2021-06-03
Criticality information on json output


rslhdyt:
I noticed the information of criticality on JSON output is missing, but when I generate the output to text the criticality info is present.

Am I missing something options or configuration?

```
{
     "type":"unpatched_gem",
     "gem":{
        "name":"actionpack",
        "version":"5.1.7"
     },
     "advisory":{
        "path":"/Users/sleekr/.local/share/ruby-advisory-db/gems/actionpack/CVE-2021-22885.yml",
        "id":"CVE-2021-22885",
        "url":"https://groups.google.com/g/rubyonrails-security/c/NiQl-48cXYI",
        "title":"Possible Information Disclosure / Unintended Method Execution in Action Pack",
        "date":"2021-05-05",
        "description":"There is a possible information disclosure / unintended method execution\nvulnerability in Action Pack which has been assigned the CVE identifier\nCVE-2021-22885.\n\nVersions Affected:  >= 2.0.0.\nNot affected:       < 2.0.0.\nFixed Versions:     6.1.3.2, 6.0.3.7, 5.2.4.6, 5.2.6\n\nImpact\n------\nThere is a possible information disclosure / unintended method execution\nvulnerability in Action Pack when using the `redirect_to` or `polymorphic_url`\nhelper with untrusted user input.\n\nVulnerable code will look like this:\n\n```\nredirect_to(params[:some_param])\n```\n\nAll users running an affected release should either upgrade or use one of the\nworkarounds immediately.\n\nWorkarounds\n-----------\nTo work around this problem, it is recommended to use an allow list for valid\nparameters passed from the user.  For example:\n\n```\nprivate def check(param)\n  case param\n  when \"valid\"\n    param\n  else\n    \"/\"\n  end\nend\n\ndef index\n  redirect_to(check(params[:some_param]))\nend\n```\n\nOr force the user input to be cast to a string like this:\n\n```\ndef index\n  redirect_to(params[:some_param].to_s)\nend\n```\n",
        "cvss_v2":null,
        "cvss_v3":null,
        "cve":"2021-22885",
        "osvdb":null,
        "ghsa":null,
        "unaffected_versions":[
           "< 2.0.0"
        ],
        "patched_versions":[
           "~> 5.2.4.6",
           "~> 5.2.6",
           "~> 6.0.3.7",
           ">= 6.1.3.2"
        ]
     }
}
```

postmodern:
[...]

PR #310, opened 2021-06-08
Add criticality level in JSON format

rslhdyt:
Currently, the criticality information is missing when tried to generate the report in JSON format.

I've updated the format/json.rb to show criticality level.

fixes: #308

### DIFF
--- a/lib/bundler/audit/advisory.rb
+++ b/lib/bundler/audit/advisory.rb
@@ -200,6 +200,11 @@ module Bundler
         id == other.id
       end
 
+      #
+      # Converts the advisory to a Hash.
+      #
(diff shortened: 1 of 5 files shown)


postmodern:
I think it would be easier to override `#to_h` in `Advisory` and `merge` in the `criticality` field from `super`.

rslhdyt:
@postmodern thanks for your feedback, I have changed the code.

postmodern:
@rslhdyt and going to ask you to add YARD documentation to `Advisory#to_h` and add a small spec for it (testing whether `criticality` is in the returned hash is sufficient).

PR #314, opened 2021-07-26
format: adding a Junit presenter format inspired by #206

paulRbr:
This is an addition of a new 'junit' format as a follow up to the
closed #206 PR.

I've mainly taken inspiration from the original template of @salzig
and adapted to the new `Formats` module.

### DIFF
--- a/lib/bundler/audit/cli/formats/junit.rb
+++ b/lib/bundler/audit/cli/formats/junit.rb
@@ -1,0 +1,122 @@
+#
+# Copyright (c) 2013-2021 Hal Brodigan (postmodern.mod3 at gmail.com)
+#
+# bundler-audit is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# bundler-audit is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with bundler-audit.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+require 'thor'
+
+module Bundler
+  module Audit
+    class CLI < ::Thor
+      module Formats
+        module Junit
+          #
+          # Prints any findings as an XML junit report.
+          #
+          # @param [Report] report
+          #   The results from the {Scanner}.
+          #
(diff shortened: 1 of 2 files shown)


postmodern:
FYI this could be it's own gem. As long as the gem is installed and provides a `lib/bundler/audit/cli/formats/junit.rb` file, `--format unit` should require the file and find the registered `:junit` format.

lib/bundler/audit/cli/formats/junit.rb
```diff
@@ -0,0 +1,118 @@
...
+            when Results::InsecureSource
+              say <<~XML
+                  <testcase id="#{ result.source }" name="Insecure Source URI found: #{result.source}">
+                    <failure message="Insecure Source URI found: #{result.source}" type="Unknown"></failure>
+                  </testcase>
+              XML
+            when Results::UnpatchedGem
+              say <<~XML
+                  <testcase id="#{ result.gem.name }" name="#{ bundle_title(result) }">
+                    <failure message="#{ result.advisory.title }" type="#{ result.advisory.criticality }">
```
postmodern, line 75: Text fields should probably be wrapped in `CGI.escapeHTML()` to prevent any quote escaping or other invalid XML characters getting into the XML.

postmodern: Probably could define a `h` or `xml` method and escape any data that goes into the XML, just in case.

lib/bundler/audit/cli/formats/junit.rb
```diff
@@ -0,0 +1,118 @@
...
+            end
+          end
+
+          def bundle_title(result)
+            "#{advisory_criticality(result.advisory).upcase} #{result.gem.name}(#{result.gem.version}) #{CGI.escapeHTML(result.advisory.title)}"
+          end
+
+          def advisory_solution(advisory)
+            unless advisory.patched_versions.empty?
+              "upgrade to #{advisory.patched_versions.join(', ')}"
```
postmodern, line 95: `Advisory#patched_versions` will definitely contain the `>` character. Should be XML escaped.

postmodern (changes requested):
Would prefer that all XML String interpolations be wrapped in a `xml(...)` escaping method that does `CGI.escapeHTML(string.to_s)`.

paulRbr:
> Would prefer that all XML String interpolations be wrapped in a `xml(...)` escaping method that does `CGI.escapeHTML(string.to_s)`.

Thanks for your review @postmodern. I've done what you suggested. And thanks for the tip about the ruby gem. 

The junit format is simple enough that I don't really feel like creating a dedicated gem just for that. But if you don't think it's going to land in `bundler-audit` let me know :slightly_smiling_face:

PR #323, opened 2022-01-10
Add Ruby 3.1 to CI

petergoldstein:
In addition to the CI configuration change this PR makes the following changes:

* Fixes YAML load in specs to permit Date
* Limits number of commit logs passed to Time.parse in spec so the string is less than 128 characters
* Adds the MFA requirement to Rubygems publishing to satisfy Rubocop
* A few formatting changes to satisfy Rubocop

### DIFF
--- a/bundler-audit.gemspec
+++ b/bundler-audit.gemspec
@@ -21,7 +21,8 @@ Gem::Specification.new do |gem|
 
   glob = lambda { |patterns| gem.files & Dir[*patterns] }
 
-  gem.files = if gemspec['files'] then glob[gemspec['files']]
+  gem.files = if gemspec['files']
+                glob[gemspec['files']]
               else
                 `git ls-files`.split($/)
               end
--- a/bundler-audit.gemspec
+++ b/bundler-audit.gemspec
@@ -22,7 +22,8 @@ Gem::Specification.new do |gem|
   glob = lambda { |patterns| gem.files & Dir[*patterns] }
 
   gem.files = if gemspec['files'] then glob[gemspec['files']]
-              else                     `git ls-files`.split($/)
+              else
+                `git ls-files`.split($/)
               end
 
   gem.executables = gemspec.fetch('executables') do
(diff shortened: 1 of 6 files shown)


bundler-audit.gemspec
```diff
@@ -22,7 +22,8 @@ Gem::Specification.new do |gem|
   glob = lambda { |patterns| gem.files & Dir[*patterns] }
 
   gem.files = if gemspec['files'] then glob[gemspec['files']]
-              else                     `git ls-files`.split($/)
+              else
+                `git ls-files`.split($/)
```
postmodern, line 26: I actually prefer the `if ... then ...` syntax. If the `else` clause is going to be on the next line, then I feel like the same should be done to the `if` clause. Or, you could just disable that rubocop rule.

lib/bundler/audit/cli.rb
```diff
@@ -76,7 +76,8 @@ def check(dir=Dir.pwd)
         report = scanner.report(ignore: options.ignore)
 
         output = if options[:output] then File.new(options[:output],'w')
-                 else                     $stdout
+                 else
+                   $stdout
```
postmodern, line 80: Ditto

lib/bundler/audit/scanner.rb
```diff
@@ -218,7 +218,8 @@ def scan_specs(options={})
         return enum_for(__method__,options) unless block_given?
 
         ignore = if options[:ignore] then Set.new(options[:ignore])
-                 else                     config.ignore
+                 else
+                   config.ignore
```
postmodern, line 222: Ditto.

bundler-audit.gemspec
postmodern, line 26: I actually prefer the `if ... then ...` syntax. If the `else` clause is going to be on the next line, then I feel like the same should be done to the `if` clause. Or, you could just disable that rubocop rule.

lib/bundler/audit/cli.rb
postmodern, line 80: Ditto.

postmodern (changes requested):
Minor stylistic feedback on the lints.

postmodern:
Looks good, only have some minor stylistic feedback on the lint changes.

petergoldstein:
@postmodern Let me know if the just committed change addresses your lint concerns.

PR #342, opened 2022-05-18
fix: don't include test files in gem pkg

NickLaMuro:
Spec files include Gemfiles with known vulnerabilities in them (on purpose), but if scanned by a naive scanner (that is looking for any Gemfile in the system), this can cause false positives.

In addition to not including this gem in production systems, we can also just remove the specs from being included in the gem package in the first place.

How to validate
---------------

Locally, I built the project, and ran `rake build` to build the gem.

Afterwards, I ran the following to determine which files were included:

```console
$ tar -xOzf pkg/bundler-audit-0.9.0.1.gem data.tar.gz | tar -tzf -
```

And validated that the `spec/` dir was included in the list.  I committed the changes for this PR, and then had the following result:

```console
$ tar -xOzf pkg/bundler-audit-0.9.0.1.gem data.tar.gz | tar -tzf -
COPYING.txt
ChangeLog.md
README.md
bin/bundle-audit
bin/bundler-audit
bundler-audit.gemspec
gemspec.yml
lib/bundler/audit.rb
lib/bundler/audit/advisory.rb
lib/bundler/audit/cli.rb
lib/bundler/audit/cli/formats.rb
lib/bundler/audit/cli/formats/json.rb
lib/bundler/audit/cli/formats/junit.rb
lib/bundler/audit/cli/formats/text.rb
lib/bundler/audit/cli/thor_ext/shell/basic/say_error.rb
lib/bundler/audit/configuration.rb
lib/bundler/audit/database.rb
lib/bundler/audit/report.rb
lib/bundler/audit/results.rb
lib/bundler/audit/results/insecure_source.rb
lib/bundler/audit/results/result.rb
lib/bundler/audit/results/unpatched_gem.rb
lib/bundler/audit/scanner.rb
lib/bundler/audit/task.rb
lib/bundler/audit/version.rb
```

### DIFF
--- a/bundler-audit.gemspec
+++ b/bundler-audit.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |gem|
   gem.files = if gemspec['files']
                 glob[gemspec['files']]
               else
-                `git ls-files`.split($/)
+                Dir.glob(%w[bin/**/* lib/**/* bundler-audit.gemspec ChangeLog.md COPYING.txt gemspec.yml README.md])
               end
 
   gem.executables = gemspec.fetch('executables') do
@@ -33,7 +33,7 @@ Gem::Specification.new do |gem|
   gem.default_executable = gem.executables.first if Gem::VERSION < '1.7.'
 
   gem.extensions       = glob[gemspec['extensions'] || 'ext/**/extconf.rb']
-  gem.test_files       = glob[gemspec['test_files'] || '{test/{**/}*_test.rb']
+  gem.test_files       = glob[gemspec['test_files'] || '{spec/{**/}*_spec.rb']
   gem.extra_rdoc_files = glob[gemspec['extra_doc_files'] || '*.{txt,md}']
 
   gem.require_paths = Array(gemspec.fetch('require_paths') {


bundler-audit.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |gem|
   gem.files = if gemspec['files']
                 glob[gemspec['files']]
               else
-                `git ls-files`.split($/)
+                Dir.glob(%w[bin/**/* lib/**/* bundler-audit.gemspec ChangeLog.md COPYING.txt gemspec.yml README.md])
```
NickLaMuro, line 27: I went with the `Dir.glob` approach out of personal preference, but if `git ls-files` is preferred, I can change this to use that instead.

postmodern: Yes `git ls-files` is preferred as that prevents non-git-tracked files from sneaking into releases.

NickLaMuro:
@postmodern Thanks for taking a look!

> The problem seems to be naive scanners that do not automatically ignore test or spec directories.

yeah, in this case, I believe it was "AWS Inspector" that was just looking around for any `Gemfile.lock` and the one in `spec` came up.  Obviously this can be fixed by just not including development gems in prod builds, but thought I would also bring this up as a proposal since a different security tool is flagging another security tool.

> Including the test files in the gemspec is generally a good practice...

I think I disagree with this notion, personally, as I see a gems as a end product, not test against.  Tests aren't included tests with something like `libpq-dev`, for example, and if tests need to be run against it, you would go back to the source code to do it.

I understand your argument, but at the end of the day, don't agree.  However, I mostly expected that, and that is totally fine, so unless you change your mind, feel free to close.

(oh, you did 😄 )

postmodern:
> I believe it was "AWS Inspector" that was just looking around for any Gemfile.lock and the one in spec came up

I would recommend submitting a bug/ticket to the AWS Inspector project about that. It should automatically ignore the `test/` and `spec/` directories as that code is never loaded/ran in production. AWS probably has plenty of engineers supporting Inspector and ignoring a directory seems like a simple fix on their end.

PR #356, opened 2022-10-19
Remove platform duplicates

ylecuyer:
Hello, lately we noticed some gems having duplicates in the report because of the various platform.

Here is an example with nokogiri:

```
    nokogiri (1.13.6)
      mini_portile2 (~> 2.8.0)
      racc (~> 1.4)
    nokogiri (1.13.6-x86_64-darwin)
      racc (~> 1.4)
    nokogiri (1.13.6-x86_64-linux)
      racc (~> 1.4)
```

And this is causing the duplicates.

As a fix, I suggest to filter and keep only the ruby platform for the check.


spec/bundle/secure/Gemfile
```diff
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
 
-gem 'rails', '~> 5.2'
+gem 'rails', '~> 7.0.4'
```
ylecuyer, line 3: need after changing the fixed commit of the advisory db

spec/spec_helper.rb
```diff
@@ -12,7 +12,7 @@ module Fixtures
   module Database
     PATH = File.join(ROOT,'database')
 
-    COMMIT = '89cdde9a725bb6f8a483bca97c5da344e060ac61'
+    COMMIT = '137a425b9f4f30f895df8765b0e773400170803d'
```
ylecuyer, line 15: Needed to get latest advisory for nokogiri

spec/scanner_spec.rb
```diff
@@ -67,7 +67,6 @@
       subject { super().scan.to_a }
 
       it "should print nothing when everything is fine" do
-        puts subject.inspect
```
postmodern: This could totally be a separate PR that I could instantly merge.

spec/bundle/unpatched_gems/Gemfile.lock
```diff
@@ -17,7 +17,14 @@ GEM
     concurrent-ruby (1.1.7)
     i18n (0.9.5)
       concurrent-ruby (~> 1.0)
+    mini_portile2 (2.8.0)
     multi_json (1.15.0)
+    nokogiri (1.13.6)
+      mini_portile2 (~> 2.8.0)
+      racc (~> 1.4)
+    nokogiri (1.13.6-x86_64-linux)
+      racc (~> 1.4)
```
postmodern, line 26: If we are going to test `Gemfile.lock` containing gems from multiple platforms or gems which explicit platforms set in the `Gemfile`, I think that should be a separate `spec/bundle/` directory (ex: `spec/bundle/unpatched_multi_platform/`).

spec/bundle/secure/Gemfile
postmodern: If we need to bump the `rails` version and the `ruby-advisory-db` commit, that can be done in a separate PR which I can immediately merge.

postmodern:
@ylecuyer maybe a better fix would be to print the gem's platform (if set) along with the gem name and version in the report? I think it would be bad if `bundler-audit` ignored an insecure gem in the `Gemfile.lock` because it's platform differs from the user's current ruby platform. This means `bundle-audit` would ensure the gems were secure for the user running it, but maybe not their coworker who's on a different ruby platform. By including the platform of the gem in the final report, we're not losing any information.

Also, under normal circumstances, you shouldn't have gems from multiple platforms in the `Gemfile.lock`.

```
source 'https://rubygems.org/'

gem 'nokogiri-ext'
gem 'nokogiri'

group :development do
  gem 'rake'
end
```

```
GEM
  remote: https://rubygems.org/
  specs:
    nokogiri (1.13.9-x86_64-linux)
      racc (~> 1.4)
    nokogiri-ext (0.1.0)
      nokogiri (~> 1.0)
    racc (1.6.0)
    rake (13.0.6)

PLATFORMS
  x86_64-linux

DEPENDENCIES
  nokogiri
  nokogiri-ext
  rake

BUNDLED WITH
   2.3.17
```

I suspect this is due to gems having explicit `platform:`/`platforms:` set in the `Gemfile` or multiple developers running `bundle install`/`bundle update` on different platforms then committing the `Gemfile.lock`?

postmodern (changes requested):
Some of these commits could be separate PRs. Also, I'm afraid that by filtering out other non-`ruby` platforms, `bundler-audit` might miss information vulnerabilities.

PR #376, opened 2023-02-10
Don't double build Dependabot PRs

nschonni:
Since they're all done through PRs, the Dependabot branch pushes don't need to be built on top


jasnow:
Out-of-date - see  PR #424/Closing PR.

PR #391, opened 2024-01-03
Fix licenses spec

viktorianer:
As listed in https://spdx.org/licenses/, the license identifier `GPL-3.0+` is deprecated. 

I replaced it with `GPL-3.0-or-later`, in order to follow the SPDX specification.
Also, I changed the `license` key in `gemspec.yml`, to `licenses`, in order to match [naming in gem specification](https://guides.rubygems.org/specification-reference/).

### DIFF
--- a/gemspec.yml
+++ b/gemspec.yml
@@ -1,7 +1,7 @@
 name: bundler-audit
 summary: Patch-level verification for Bundler
 description: bundler-audit provides patch-level verification for Bundled apps.
-license: GPL-3.0+
+license: GPL-3.0-or-later
 authors: Postmodern
 email: postmodern.mod3@gmail.com
 homepage: https://github.com/rubysec/bundler-audit#readme


gemspec.yml
```diff
@@ -1,7 +1,7 @@
 name: bundler-audit
 summary: Patch-level verification for Bundler
 description: bundler-audit provides patch-level verification for Bundled apps.
-license: GPL-3.0-or-later
+licenses: GPL-3.0-or-later
```
postmodern, line 4: If the gem only has one license, why use `licenses`?

postmodern (commented):
I'll accept the license identifier change, but I'm curious why use `licenses` if the gem only has one license? One of the reasons I use `gemspec.yml` is to abstract away a lot of the quirks of rubygems `Gem::Specification`.

gemspec.yml
viktorianer: But then, why you use `authors`, when here is also only one? I think consistency to gem specification make sens here. That's why `licenses` and `authors`.

viktorianer: OK, I reverted `licenses` change.

postmodern (approved):

PR #397, opened 2024-10-27
Remove Dir.chdir for thread-safety

### LINKED ISSUE #156: Bundler audit is not thread safe
Because of the Dir.chdir in lib/bundler/audit/database.rb, the gem itself is not threadsafe

lovro-bikic:
Currently, `Bundler::Audit::Database#update!`, `#commit_id` and `#last_updated_at` fail in multi-threaded environments (like background processing jobs) because they use `Dir.chdir`.

[From Ruby docs](https://docs.ruby-lang.org/en/3.3/Dir.html#method-c-chdir) on `Dir.chdir`:
> In a multi-threaded program an error is raised if a thread attempts to open a chdir block while another thread has one open, or a call to chdir without a block occurs inside a block passed to chdir (even in the same thread).

This PR removes the need for `Dir.chdir` by using [`Open3.capture2`](https://docs.ruby-lang.org/en/3.3/Open3.html#method-i-capture2) from the stdlib with the [`chdir` option](https://docs.ruby-lang.org/en/3.3/Process.html#module-Process-label-Working+Directory+-28-3Achdir-29).

External behavior should stay the same: `Dir.chdir` raises an `Errno::ENOENT` error given an non-existent directory, and the same holds with the new implementation.

Fixes https://github.com/rubysec/bundler-audit/issues/156

### DIFF
--- a/lib/bundler/audit/database.rb
+++ b/lib/bundler/audit/database.rb
@@ -17,6 +17,7 @@
 
 require 'bundler/audit/advisory'
 
+require 'open3'
 require 'time'
 require 'yaml'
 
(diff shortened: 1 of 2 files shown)


simi:
@lovro-bikic what about to use `git -C` (work dir), would that work also? 🤔

lovro-bikic:
Nice one! I believe we could use it. I only have two questions:
1. `-C` option was introduced in [v1.8.5](https://lwn.net/Articles/575726/). I'm not sure if we should support older Git versions?
2. While undocumented and untested, the gem currently raises `Errno::ENOENT` for unknown directories. With `git -C`, the error will look like below. Is that acceptable?

```
# with fake DB path "blabla"

> bin/bundler-audit update
Updating ruby-advisory-db ...
fatal: cannot change to 'blabla': No such file or directory
```

PR #401, opened 2025-05-02
Fix different database path used for `check` command

oz-tal:
If the database path is overridden with the environment variable `BUNDLER_AUDIT_DB`,  the `check` command doesn't take it into account, while `download`, `update` and `stats` do.

This fix ensure `check` use the same method as the other commands to find the default path to the database.


postmodern:
Ended up using `Database::DEFAULT_PATH` instead to avoid an extra method-call. 0d870299d19f3ff76d1026ed6358be9261eb4f31

Should be released in 0.9.3.

PR #402, opened 2025-05-13
Only load from `cgi` what is required for Ruby 3.5

Earlopain:
In Ruby 3.5 most of the `cgi` gem will be removed. Only the various escape/unescape methods will be retained by default.

But: `cgi/escape` only exists since Ruby 2.3. So I fall back to `cgi/util` on earlier versions.

https://bugs.ruby-lang.org/issues/21258


postmodern:
The linked upstream issue is still open and they appear to still be discussing what to do with `cgi`. Testing under ruby-3.5.0-preview1, one can still require `cgi`, `cgi/escape`, and `cgi/util`, and call `CGI.escapeHTML`. This doesn't appear to be an issue, yet. Feel free to re-open when `cgi` is removed from stdlib.

postmodern:
Hmm, it looks like hsbt's PR was merged into Ruby master, so `cgi` will be soft-removed in 3.5.0-preview2, with a placeholder `cgi.rb` file left behind recommending using `cgi/escape` instead. At that point I'll probably use this code from the PR instead:

```ruby
      begin
        require 'cgi/escape'
      rescue LoadError
        require 'cgi/util'
      end
```

Earlopain:
Sure, I changed it and left a comment for which ruby version it applies

postmodern:
I have tested bundler-audit 0.9.2 and the git `master` version (0.9.3) and they both seem to work under Ruby 3.5 and 4.0.

`Gemfile`:
```ruby
source 'https://rubygems.org/'

gem 'bundler-audit'
gem 'rake', '< 12.3.3' # CVE-2020-8130
```
```
$ bundle i
$ bundle exec bundler-audit check --format=junit
<?xml version="1.0" encoding="UTF-8" ?>
<testsuites id="1764373267" name="Bundle Audit">
  <testsuite id="Gemfile" name="Ruby Gemfile" failures="1">
    <testcase id="rake" name="HIGH rake(12.3.2) OS Command Injection in Rake">
      <failure message="OS Command Injection in Rake" type="high">
        Name: rake
        Version: 12.3.2
        Advisory: CVE-2020-8130 GHSA-jppv-gw3r-w3q8
        Criticality: High
        URL: https://github.com/advisories/GHSA-jppv-gw3r-w3q8
        Title: OS Command Injection in Rake
        Solution: update to &#39;&gt;= 12.3.3&#39;
      </failure>
    </testcase>
  </testsuite>
</testsuites>
```

Even manually requiring `cgi` still works under both Ruby versions.

```
$ chruby ruby-3.5
$ ruby -r cgi -e 'puts CGI.escapeHTML("<>")'
&lt;&gt;
$ chruby ruby-4.0
$ ruby -r cgi -e 'puts CGI.escapeHTML("<>")'
&lt;&gt;
```

It also appears from the ruby-core bug discussion, that they will keep `cgi` in stdlib a bit longer but with deprecation warnings. Unless I am missing something, this appears to be more of a micro-optimization, and not a critical bug fix?

When `cgi` *is* finally removed from stdlib, we will need to add the `cgi` gem as a dependency or vendor the `escapeHTML` method.

Earlopain:
You need to run with warnings, otherwise you will not see this. You will find that most of the actual cgi functionality is gone. `cgi/cookie`, `cgi/session`, etc (for those few that actually use them).

```
$ RBENV_VERSION=ruby-dev RUBYOPT=-w ruby -rcgi -ve ""
ruby 4.0.0dev (2025-11-29T02:06:53Z master 82b91ec7e5) +PRISM [x86_64-linux]
CGI library is removed from Ruby 4.0. Please use cgi/escape instead for CGI.escape and CGI.unescape features.
If you need to use the full features of CGI library, Please install cgi gem.

$ RBENV_VERSION=ruby-dev RUBYOPT=-w ruby -rcgi/cookie -ve ""
ruby 4.0.0dev (2025-11-29T02:06:53Z master 82b91ec7e5) +PRISM [x86_64-linux]
<internal:/home/user/.rbenv/versions/ruby-dev/lib/ruby/4.0.0+0/rubygems/core_ext/kernel_require.rb>:139:in 'Kernel#require': cannot load such file -- cgi/cookie (LoadError)
```

It is not a "critical" butfix but gets rid of the warning (when junit is used).

Upstream rails tests run with warnings enabled by default, so I checked dependencies for `require "cgi"` and updated them. I don't think they use the junit reporter so it should not matter for that specific case, but what is the reason for rejecting this? It is what ruby itself recommends.

PR #407, opened 2025-10-22
Documentations fixes

biow0lf:


postmodern:
I prefer explicit `return`s and any README example that contains output should not have syntax highlighting on it. I went ahead and fixed the documentation typo and converted those two indented examples to use markdown code fences.

PR #395, opened 2024-09-12
Remove tests files and other config-related files from the gem packages

kpumuk:
Current rubygems guidelines suggest to only include runtime files as a best practice: https://github.com/rubygems/guides/issues/90. Additionally, the test-files contain insecure `Gemfile.lock` files which can sometimes trigger other vulnerability scanners when they scan `bundler-audit`.

In this MR I propose to remove the following:

* files starting with `.` in the root of the repository (includes GitHub workflows and issue templates, RSpec configuration, Yard settings)
* `Gemfile` as it is only needed for development and not used in runtime
* `gemspec.yml` which is used to construct gemspec and not used after that
* `spec/` includes test files and fixtures

To consider:

* `Rakefile` references `rspec` and `yard` which are development-only dependencies.
* `bundler-audit.gemspec` since it is only used in development to produce metadata, added to the gem package as a YAML-serialized `Gem::Specification`
* ~~`ChangeLog.md` probably is not looked at ever and only takes space.~~

## Prior art

Since RuboCop added `test_files` to deprecated gemspec attributes, the directive was removed in https://github.com/rubysec/bundler-audit/pull/370 (see discussion in https://github.com/rubocop/rubocop/issues/10675).

Another pull request that addresses a similar concern is https://github.com/rubysec/bundler-audit/pull/416.

## File list difference

To test, I built the gem with `rake build` and compared the listing before and after the change:

```bash
tar -xOzf pkg/bundler-audit-0.9.2.gem data.tar.gz | tar -tzf -
```

Full diff:

```diff
--- before.txt	2024-09-12 06:48:59
+++ after.txt	2024-09-12 06:48:38
@@ -1,21 +1,10 @@
-.document
-.github/FUNDING.yml
-.github/ISSUE_TEMPLATE/bug-report.md
-.github/ISSUE_TEMPLATE/feature-request.md
-.github/workflows/ruby.yml
-.gitignore
-.rspec
-.rubocop.yml
-.yardopts
 COPYING.txt
 ChangeLog.md
-Gemfile
 README.md
 Rakefile
 bin/bundle-audit
 bin/bundler-audit
 bundler-audit.gemspec
-gemspec.yml
 lib/bundler/audit.rb
 lib/bundler/audit/advisory.rb
 lib/bundler/audit/cli.rb
@@ -34,37 +23,3 @@
 lib/bundler/audit/scanner.rb
 lib/bundler/audit/task.rb
 lib/bundler/audit/version.rb
-spec/advisory_spec.rb
[...]

PR #416, opened 2026-03-18
exclude spec/ directory from published gem

neumayr:
Hey @postmodern 

Thank you for the quick feedback. I've addressed your comments.

- exclude `spec/` directory from published gem
- follow up of #415


simi (approved):

simi:
@jasnow we can start here if https://github.com/rubysec/bundler-audit/pull/395 is too aggressive for now.

jasnow:
> @jasnow we can start here if #395 is too aggressive for now.

Will defer to @flavorjones and you.

flavorjones:
Closed in favor of #395

PR #419, opened 2026-04-02
Add `exclude` option to skip gems by name during audit

mlarraz:
## Description

Adds an `exclude` configuration option (and `--exclude` / `-e` CLI flag) that skips entire gems by name during scanning, so they never hit the advisory database at all.

This is semantically different from `ignore` (which skips specific advisory IDs after lookup): `exclude` is a scoping decision — "this gem isn't part of my application" — while `ignore` is a risk acceptance — "I know about this CVE and accept it."

Useful for Rails apps that bundle gems they don't actually use (e.g. `activestorage`, `actiontext`, `actionmailbox`) where every new CVE triggers a false audit failure. Currently the only workaround is adding each CVE ID to the `ignore` list, which is tedious and never-ending.

## Usage

**Config file (`.bundler-audit.yml`):**
```yaml
---
ignore:
  - CVE-2024-1234
exclude:
  - activestorage
  - actiontext
```

**CLI:**
```
bundle-audit check --exclude activestorage actiontext
```

## Changes

- **`Configuration`**: new `attr_reader :exclude` (returns `Set<String>`), YAML validation mirroring `ignore`
- **`Scanner#scan_specs`**: resolves `exclude` from options or config, skips gem before advisory lookup via `next if exclude.include?(gem.name)`
- **`CLI#check`**: new `--exclude` / `-e` array option, passed through to `scanner.report`
- **Tests**: validation specs, initializer specs, scanner specs (option-based and config-file-based), new fixtures

## Test plan

- [x] New configuration validation tests (not an array, contains non-string)
- [x] New `#initialize` tests (default empty Set, given `:exclude` list)
- [x] Scanner test: `:exclude` option filters gems from results
- [x] Scanner test: exclude from `.bundler-audit.yml` config file works
- [x] All existing tests continue to pass

### DIFF
--- a/lib/bundler/audit/cli.rb
+++ b/lib/bundler/audit/cli.rb
@@ -37,6 +37,7 @@ module Bundler
       method_option :quiet, type: :boolean, aliases: '-q'
       method_option :verbose, type: :boolean, aliases: '-v'
       method_option :ignore, type: :array, aliases: '-i'
+      method_option :exclude, type: :array, aliases: '-e'
       method_option :update, type: :boolean, aliases: '-u'
       method_option :database, type: :string, aliases: '-D',
                                default: Database::DEFAULT_PATH
@@ -73,7 +74,7 @@ module Bundler
                      exit 1
                    end
 
-        report = scanner.report(ignore: options.ignore)
+        report = scanner.report(ignore: options.ignore, exclude: options.exclude)
 
         output = if options[:output]
                    File.new(options[:output],'w')
--- a/lib/bundler/audit/configuration.rb
+++ b/lib/bundler/audit/configuration.rb
@@ -77,6 +77,16 @@ module Bundler
             end
 
             config[:ignore] = value.children.map(&:value)
+          when 'exclude'
+            unless value.is_a?(YAML::Nodes::Sequence)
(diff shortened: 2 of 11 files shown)


jonswar (approved):

flavorjones:
Rebased onto current `master` to pull in CI matrix changes.

flavorjones (changes requested):
This seems reasonable given Rails allows applications to opt out of using gems via options like `--skip-active-storage` but those gems are still hard dependencies in `Gemfile.lock`.

I'm open to merging this. But please take a look at the failing test, seems like there's a missing `Gemfile.lock` in one of the test fixtures.

mlarraz:
> This seems reasonable given Rails allows applications to opt out of using gems via options like `--skip-active-storage` but those gems are still hard dependencies in `Gemfile.lock`.
> 
> I'm open to merging this. But please take a look at the failing test, seems like there's a missing `Gemfile.lock` in one of the test fixtures.

Thanks, looks like it needed a `git add -f`. Lockfiles are in .gitignore but needed for CI to pass.

flavorjones (approved):
@simi Can you review, and if this seems good to you, I'm OK with merging.

jonswar:
Any blockers to this being merged?

Issue #309, opened 2021-06-04
Add a Security Policy


postmodern:
Add a `SECURITY.md` file explaining how to report vulnerabilities in bundler-audit.

* Which email address should they be sent to? (rubysec's mailing list or my email addres?)
* Which PGP key, if any, should be used to encrypt emails? (I can volunteer my PGP pubkey)

reedloden:
I'm a bit biased here due to it being my employer (and the fact that I manage this particular offering), but HackerOne offers a [completely free version](https://www.hackerone.com/company/open-source-community) for open source projects. Might I suggest that as an alternative to email and PGP? Ruby, Rails, and RubyGems all use it already, just as examples.

kallal79:
HI @postmodern  , I’d like to work on this issue, please assign it to me.

PR #420, opened 2026-04-03
Add SECURITY.md vulnerability reporting policy

kallal79:
Implement issue #309 by adding a dedicated security policy for bundler-audit.

Changes:

- Add SECURITY.md with private vulnerability reporting instructions.

- Document email reporting channel and disclosure expectations.

- Add spec/security_policy_spec.rb to verify policy file presence and core guidance.

This commit intentionally contains only security-policy related files.


SECURITY.md
```diff
@@ -0,0 +1,32 @@
+# Security Policy
+
+If you discover a security vulnerability in bundler-audit, please report it
+privately.
+
+## Reporting A Vulnerability
+
+- Email: postmodern.mod3@gmail.com
```
simi: I assume we need to change the email now. Any idea where to point now? @jasnow what about your contact?

jasnow: Opened this yesterday for `security@rubysec.com`: 
 * https://github.com/rubysec/bundler-audit/issues/428
I just created a backup gmail email and will share it if we need it.
 * FYI: rubysec@gmail.com is taken so I sent it an email asking who they were.

spec/security_policy_spec.rb
```diff
@@ -17,6 +17,6 @@
   end
 
   it 'contains a security contact channel' do
-    expect(content).to include('postmodern.mod3@gmail.com')
+    expect(content).to include('security@rubysec.com')
```
jasnow: Do you own the security@rubysec.com email? I don't.

SECURITY.md
```diff
@@ -5,7 +5,7 @@ privately.
 
 ## Reporting A Vulnerability
 
-- Email: postmodern.mod3@gmail.com
+- Email: security@rubysec.com
```
jasnow: Do you own the security@rubysec.com email? I don't.

SECURITY.md
flavorjones: I'd like to propose that we use a contact address that is itself a resource with shared ownership (e.g., a private Google group where we can manage membership and permissions), presuming we gain control of gem publishing.

jasnow: > resource with shared ownership (e.g., a private Google group where we can manage
>  membership and permissions), presuming we gain control of gem publishing.
My previous thoughts are in ISS #428 but I can support @flavorjones' approach too.

kallal79:
FIX

PR #446, opened 2026-07-10
Add support for expiring advisory ignores

abeljim8am:
Allow advisory ignores in `.bundler-audit.yml` to expire automatically, so temporary vulnerability exceptions do not remain in place indefinitely.

Timed ignores use an advisory ID and an inclusive ISO 8601 `until` date:

```yaml
ignore:
  - id: CVE-YYYY-XXXX
    until: 2026-08-31
```

Existing string entries continue to be ignored indefinitely. Once a timed ignore expires, the advisory is reported normally. Invalid dates, missing fields, and unknown keys are rejected when loading the configuration.

The configuration and scanner specs cover active and expired ignores, backward compatibility, and validation failures.


simi:
WDYT @jasnow?

jasnow:
I'm basically unsure about why ignores are needed. We have at least 3 customers: "bundle audit" users, database scrapers, and bad guys. The last two will not ignore them and the first one can deal with it locally with GitHub action/bash aliases/etc.
Feels more like a convenience feature than must-have.

On the other hand, it probably will not hurt to have.

simi:
I assume it is feature to users to ignore given reports like for next 14 days or so to get some time to resolve and get only new reports. But just blindly guessing. @abeljim8am can you explain more?

abeljim8am:
Sometimes we hit advisories where downstream fixes aren't available yet. A temporary bypass mechanism keeps these visible without blocking CI in the short term.

Right now, we maintain a wrapper script around bundle-audit for this. Native configuration support would let us simplify our setup and rely entirely on bundler-audit.

jasnow:
> Sometimes we hit advisories where downstream fixes aren't available yet. A temporary bypass mechanism keeps these visible without blocking CI in the short term.
> 
> Right now, we maintain a wrapper script around bundle-audit for this. Native configuration support would let us simplify our setup and rely entirely on bundler-audit.

I think rails has a wrapper too - @flavorjones right?

PR #447, opened 2026-07-10
Add inherit config option

jonswar:
Add `inherit` config option, allowing other config files to be merged into the current config. Similar to rubocop's [inherit_from](https://docs.rubocop.org/rubocop/latest/configuration/inheritance.html) directive. e.g.

```
---
inherit:
  - ../bundler-audit-shared.yml

ignore:
  - CVE-1
  - ...
```

This is useful for large repos with many apps. Each app has its own ignores, but occasionally there's a desire to ignore a CVE across all apps. Right now in our organization we have to add the ignore to 30+ files, which is tedious and obscures commonalities.


simi:
Why not to use `inherit_from` also? :thinking:

jonswar:
> Why not to use `inherit_from` also? 🤔

Good point, changed to `inherit_from`.

PR #118, opened 2015-10-29
Audit ruby and gems

jeffreyc:
https://github.com/rubysec/ruby-advisory-db now includes security bulletins for Rubies and libraries (i.e., RubyGems). There is a need to audit Ruby and RubyGems, and `bundler-audit` is a great candidate for this functionality, since most of the necessary functionality is already implemented. This pull request adds support for auditing the current Ruby and RubyGems versions.

The Ruby version is determined via `RUBY_ENGINE`, `RUBY_VERSION`, and `RUBY_PATCHLEVEL`. The RubyGems version is determined by shelling out to `gem -v`. This won't address environments where bundler-audit is run using a different Ruby engine (e.g., when auditing a project in a different directory), but seems consistent with https://github.com/rubysec/ruby-advisory-db/pull/45#issuecomment-32545332.

I _think_ the patchlevel issues should be straightforward. `ruby-advisory-db` seems to have standardized on one of:
- `major.minor.bugfix`
- `major.minor.bugfix.patchlevel`
- `major.minor.bugfix.preview.patchlevel`

I tested an assortment of versions, and verified that things behave as expected for:
- 1.9.3-p286
- 1.9.3-p551
- 2.2.1
- 2.2.2
- 2.2.3
- jruby-9.0.3.0
- rbx-2.5.8

It gets a little tricky with pre-release versions, like 
- 2.1.0preview1
- 2.2.0preview2
- 2.2.0rc1
- jruby-master+graal-dev

In those cases, `RUBY_PATCHLEVEL` is `-1`, even in the case of 2.2.0preview2. Since `ruby-advisory-db` is representing these as `major.minor.bugfix.preview.patchlevel`, I've worked around this by shelling out, though that's less than ideal.

Adding test coverage required updating the `ruby-advisory-db` submodule. This revealed an interesting issue with git:

```
% git log --pretty="%cd" -1
2015-10-28 12:13:14 -0500
% git log --date=iso8601 --pretty="%cd" -1
2015-10-28 12:13:14 -0400
```

To address that, I updated the `rake db:update` task to be consistent with `Bundler::Audit::Databse.path`.

For new functions, I included `@since 0.5.0`, assuming that when this is released, it will be a minor version bump.

PR #451, opened 2026-07-16
Check ruby version from lockfile for advisories

eliotsykes:
Scans the ruby version in the lockfile and reports any corresponding advisories. 

This should help projects already running bundler-audit become aware of new ruby security vulnerabilities and patches sooner.

@postmodern mentioned being open to checking the ruby version in `Gemfile.lock` sometime ago in #118 
> I'm open to checking the ruby version specified in the Gemfile.lock ... [link](https://github.com/rubysec/bundler-audit/pull/118#issuecomment-374377322).

If a scanned lockfile does not contain the `RUBY VERSION` metadata, then this ruby check is silently skipped.

Thanks in advance for any feedback.


simi:
What about the other formatters (junit, json)?

eliotsykes:
@simi Thanks for the feedback, I'll add support for the junit and json formatters.

eliotsykes:
Hi @simi, json and junit formatting support has now been added

lib/bundler/audit/cli/formats/json.rb
```diff
@@ -56,6 +56,37 @@ def criticality_label(advisory)
...
+          #
+          # Prepares a result hash before it is formatted as JSON.
+          #
+          # @param [Hash] result
+          #   A result
+          #
+          def prepare_result(result)
+            if (advisory = result[:advisory])
+              advisory.delete(:gem)
+              advisory.delete(:engine)
```
eliotsykes, line 87: This was needed to exclude these two new `Advisory` attributes from the JSON output. 

I am avoiding adding new data to the output JSON without good reason, to reduce the risk of breaking anything that depends on this JSON having the same keys it always has.

